PR #9, opened 2015-12-21
RYA-23 exposing triple pattern strategy from rya context

pujav65:


amihalik:
I think I understand what you're doing here, but I'd prefer the M/R input format read from the config file and derive the table layout rather than exposing this.

PR #10, opened 2015-12-21
RYA-21 adding input format and sample mappers/reducers

pujav65:


dao/accumulo.rya/pom.xml
```diff
@@ -66,6 +66,13 @@ under the License.
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.mrunit</groupId>
+            <artifactId>mrunit</artifactId>
+            <classifier>hadoop2</classifier>
+            <version>1.1.0</version>
```
amihalik: you should remove the version and classifier (I think ... I'm not positive about the classifier).  We define these in the root pom.

PR #15, opened 2016-01-28
RYA-32 Improve how metadata and values are written to Accumulo PCJ tables

meiercaleb:
Created a class AccumuloPcjSerializer within the rya.indexing project for serializing and deserializing binding sets for storing and reading solutions in precomputed join tables.  Also developed the class PcjTables to encapsulate the logic for creating a PCJ table.  Refactored some of the core precomputed join classes to utilize these updates and included a number of comments to make the code clearer.  Updated a large number of tests within the packages mvm.rya.indexing.external and mvm.rya.indexing.external.tupleSet to reflect these updates.


joshelser:
Hi @meiercaleb. Thanks for the pull request. It looks like you have included lots of unrelated whitespace changes: removing whitespace on empty lines and changing spaces to tabs. It would be great if you could undo those changes as they really distract from the code you actually did change :)

PR #18, opened 2016-02-16
RYA-35 - Added a class that converts BindingSet objects into Strings …

kchilton2:
…and back again.


amihalik:
Looks good.  Thanks!

PR #20, opened 2016-02-19
RYA-41 Fixed a bug where AccumuloIndexSet assumes the first node of a…

kchilton2:
… ParsedQuery is always a Projection node.


amihalik:
Looks good.  I'll merge this on Sunday unless there are other concerns.

PR #23, opened 2016-03-02
RYA-44 testMilliSecondsNoZone test case fails when running from +0530 time zone

AmilaWijayarathna:


DLotts:
Looks great!
Thanks go out to our new friend Amila  -- thanks for fixing Rya to work on both sides of the Earth!
david.

PR #25, opened 2016-03-02
RYA-7 Applying Correct License

amihalik:


joshelser:
Looks like that was the bad license headers I found. Thanks for grabbing those.

If you want to get Rya into a good place going forward, I'd recommend also getting `mvn apache-rat:check` to run without errors. I stepped through it (just removing the offending files) -- I only came across test files and META-INF/service definitions (both of which are somewhat expected). For the containing maven module, just add a pluginConfiguration entry in the pom.xml adding exclusions for the files. After this (or maybe after the first release -- so we have someone's eyes on it other than just mine), we can check off the other copyright box.

amihalik:
Thanks @joshelser, but I need a bit more help.

I don't quite understand the apache-rat behavior.  We have those exclusions in the parent pom, we have apache-rat check enabled by default, and everything works as expected if we do a `mvn clean install` (i.e. the exclusions are propagagted to the submodules).  However those exclusions aren't propagated if we do a `mvn apache-rat:check`.  Should I manually cut-and-paste those exclusions to the submodules?

Secondly, I don't understand why apache-rat didn't catch the  "Copyright Rya" in these two files.  (And I really don't understand how you caught that either :) ).  Is there a more stringent setting for apache-rat?

Thanks for your help!

joshelser:
>  We have those exclusions in the parent pom, we have apache-rat check enabled by default, and everything works as expected if we do a mvn clean install (i.e. the exclusions are propagagted to the submodules).

Oh interesting. I assume you mean that the apache-rat check is running with the `mvn clean install`? It could just be a configuration issue.

>  However those exclusions aren't propagated if we do a mvn apache-rat:check

IIRC, you should really set the configuration in each sub-module. There are defaults (such as target/) which are automatically included via the apache parent pom, but it's good practice to localize any exclusions to the rat check in the sub-modules (to avoid unexpected exclusions). I can try to help you out with this if you get stuck.

>  I don't understand why apache-rat didn't catch the "Copyright Rya" in these two files. (And I really don't understand how you caught that either :) ). Is there a more stringent setting for apache-rat?

I just grep'ed for Copyright. The license header might have been close enough to the ASL that it passed the RAT check, but I was more looking for places where you had copyrights listed (the rules are that you don't include a copyright for ASF-owned code).

amihalik:
gotcha, thanks!  I'll put in another issue to set the configuration in each sub-module.

PR #27, opened 2016-03-10
RYA-49 Updated Accumulo version from 1.6.4 to 1.6.5 in the Vagrant fi…

ejwhite922:
…le.  1.6.4 is no longer available on any mirror sites.  The script now stores the version in a variable so it only needs to be updated in one spot if the version changes again.


pujav65:
Can this be closed?

PR #34, opened 2016-03-25
updated Accumulo and Hadoop version numbers

bwmcbride:
Interim fix for the out of data Accumulo download URL.  Interim because it would be better to switch to using URLs that are known to be stable at least whilst the vagrant file refers to them.

See also pull request https://github.com/apache/incubator-rya/pull/27 (which I hadn't noticed before I did my own updates.

Also:
- adds download URL for base box
- checks that downloads work and fails the script if they don't
- uses symbolic version numbers throughout

### DIFF
--- a/extras/vagrantExample/src/main/vagrant/Vagrantfile
+++ b/extras/vagrantExample/src/main/vagrant/Vagrantfile
@@ -28,7 +28,11 @@
 
 Vagrant.configure(2) do |config|
 
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "ubuntu/trusty64"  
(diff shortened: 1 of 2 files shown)


DLotts:
Wow, this is awesome!  Love the download function.
This issue has gotten a lot of attention, as it should.  I'll have to merge my changes into yours that I did for RYA-54  https://issues.apache.org/jira/browse/RYA-54.  
I made it handle restarts, and re-provisioning.
I used the approach of adding    "|| exit 150"   to the ends of the lines that can fail.
david.

amihalik:
Thanks!  I merged this PR.  Please close it :)

PR #35, opened 2016-04-04
RYA-50 Updated Rya PCJs to support SPARQL queries with OPTIONALs.

meiercaleb:


extras/indexing/src/main/java/mvm/rya/indexing/external/tupleSet/AccumuloPcjSerializer.java
```diff
@@ -19,104 +19,198 @@
...
 	/**
-	 *
+	 * 
 	 * @param bs {@link BindingSet} to be serialized
 	 * @param varOrder order in which binding values should be written to byte array
 	 * @return byte array containing serialized values written in order indicated by varOrder
 	 * @throws RyaTypeResolverException
+	 * @deprecated Use {@link #convert(BindingSet, VariableOrder)} instead.
 	 */
+    @Deprecated
```
amihalik: Why not remove the method?

PR #41, opened 2016-05-16
FederatedRyaExample

jasonyang1686:
Some example codes using federated Rya with current Rya state.

### DIFF
--- a/extras/indexingExample/src/main/java/FederatedRyaMockExample.java
+++ b/extras/indexingExample/src/main/java/FederatedRyaMockExample.java
@@ -1,0 +1,297 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.accumulo.core.client.AccumuloException;
+import org.apache.accumulo.core.client.AccumuloSecurityException;
+import org.apache.accumulo.core.client.Connector;
+import org.apache.accumulo.core.client.TableExistsException;
+import org.apache.accumulo.core.client.TableNotFoundException;
+import org.apache.accumulo.core.client.mock.MockInstance;
+import org.apache.accumulo.core.client.security.tokens.PasswordToken;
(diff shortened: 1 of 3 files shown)


extras/indexingExample/src/main/java/FederatedRyaMockExample.java
```diff
@@ -0,0 +1,297 @@
...
+import mvm.rya.indexing.accumulo.ConfigUtils;
+import mvm.rya.indexing.accumulo.geo.GeoConstants;
+import mvm.rya.indexing.external.tupleSet.PcjTables;
+import mvm.rya.indexing.external.tupleSet.PcjTables.PcjException;
+import mvm.rya.indexing.external.tupleSet.PcjTables.PcjVarOrderFactory;
+import mvm.rya.rdftriplestore.inference.InferenceEngineException;
+import mvm.rya.sail.config.RyaSailFactory;
+
+public class FederatedRyaMockExample {
+	private static final Logger log = Logger.getLogger(RyaDirectExample.class);
```
pujav65: change the logging to reference the right class

extras/indexingExample/src/main/java/FederatedRyaRealExample.java
```diff
@@ -0,0 +1,281 @@
...
+	// Connection configuration parameters
+	//
+
+	private static final boolean USE_MOCK_INSTANCE = false;
+	private static final boolean PRINT_QUERIES = true;
+	private static final String INSTANCE = "dev";
+	private static final String RYA_TABLE_PREFIX = "rya_";
+	private static final String AUTHS = "";
+	//the second VM's IP is 192.168.33.20
+	private static final String SESAME_SERVER="http://192.168.33.20:8080/openrdf-sesame";
```
pujav65: don't check in ips

extras/indexingExample/src/main/java/FederatedRyaRealExample.java
```diff
@@ -0,0 +1,281 @@
...
+		conf.set(ConfigUtils.USE_PCJ, "false");
+		conf.set(ConfigUtils.USE_GEO, "false");
+		conf.set(ConfigUtils.USE_FREETEXT, "false");
+		conf.set(ConfigUtils.USE_TEMPORAL, "false");
+		conf.set(RdfCloudTripleStoreConfiguration.CONF_TBL_PREFIX,
+				RYA_TABLE_PREFIX);
+		conf.set(ConfigUtils.CLOUDBASE_USER, "root");
+		conf.set(ConfigUtils.CLOUDBASE_PASSWORD, "root");
+		conf.set(ConfigUtils.CLOUDBASE_INSTANCE, INSTANCE);
+		conf.set(ConfigUtils.CLOUDBASE_ZOOKEEPERS,"192.168.33.10:2181");
```
pujav65: don't check in ips

pujav65:
I'm not sure that we want to add all three of these examples.  There is a lot of duplicated code between them.  
I think that you likely would only want FederatedRyaMockExample_diff_prefix.java (though maybe rename that to FederatedRyaMockExample.java), and the sesame integration.  If you could combine the diff prefix and the sesame integration into the same example (so that they don't duplicate code anymore), I will merge this.

extras/indexingExample/src/main/java/FederatedRyaRealExample.java
DLotts: Note that this is a non-internet-routable address similar to the one used by our VM in the Rya Vagrant example.

PR #49, opened 2016-06-03
Updated PcjTables to be compatible with MockInstance

meiercaleb:
The PcjTables class now checks whether the Accumulo Connector parameter is for a MockInstance. If the Connector is for a MockInstance, the table cardinality is updated using a BatchWriter as opposed to a ConditionalWriter, which is what is used for normal Instances.  There is a risk that if the cardinality is updated asynchronously for a MockInstance the cardinality will be incorrect.  But this change is mainly to facilitate the use of PcjTables with MockInstances for testing purposes.


pujav65:
merged.

pujav65:
hey caleb can you close this pull request?

PR #53, opened 2016-06-28
Updated config to not use inferencing by default 

meiercaleb:
Updated config to not use inferencing by default.  Also fixed RyaDirectExample so that it closes all SailRepository and SailRepositoryConnection objects that get created.


pujav65:
lots of reformatting -- maybe in the future try to avoid checking that stuff in

pujav65:
merged

PR #56, opened 2016-07-07
Add Merge Tool

ejwhite922:
RYA-75
Added Merge Tool which merges a child's Accumulo instance to its parent instance after the child has collected/modified its data set. The Copy Tool allows the creation of a child instance from a parent Accumulo instance and the data that is copied can be filtered by time.  This also contains the query based copying.


amihalik:
Did you mean to check in those accumulo table files (i.e. /test/copy_tool_file_output/**)?

ejwhite922:
Yes, those are test files that can be read in.

pujav65:
@amihalik did you have any other comments?  Otherwise I was going to merge -- looked fine to me

pujav65:
hey @ejwhite922, this failed the rat check.  Can you add the license and then update the pull request?

amihalik:
@ejwhite922 going back to the binary files like:

extras/rya.merger/resources/test/copy_tool_file_output/ct_spo/files/part-r-00005.rf

I'm concerned that your tests based on these files will break if we upgrade Accumulo.  Then you'll have to regenerate them for a newer version of Accumulo.  Can you check in a text version of the data instead and generate these tables during the test?

ejwhite922:
@pujav65 I updated the files that were missing licenses so the rat check passes.

ejwhite922:
@amihalik the Copy Tool/Merge Tool relies on creating and reading .rf files so it'd be good if the tests do fail if a later version of Accumulo breaks it so we'll know about it.  But changing versions of Accumulo shouldn't affect this.

PR #57, opened 2016-07-08
RYA-130 Fixed Rya Mongo Multi Binding Set Join

amihalik:
Fixed bug in Mongo Query Engine; added test


pujav65:
good catch -- about to merge.  We should think about adding a correctness test like what Jeff suggested.

pujav65:
merged

pujav65:
@amihalik can you close this pull request?

PR #58, opened 2016-07-18
RYA-105: Fluo Integration

meiercaleb:
Integrated Fluo with Rya so that when triples are inserted into Rya via the AccumuloRyaDAO, they are also inserted into Fluo so that any registered Precomputed Joins are incrementally updated.


pujav65:
again, lots of formatting changes so hard to figure out exactly what changed.  It looks like the indexer was added to the factory, tests were fixed, and a few bugs in initialization were fixed (moving initialization to the init method, and changing to only write the version triple to the core tables).  Looked fine to me -- I will merge.

pujav65:
merged

pujav65:
@meiercaleb can you close this pull request?

PR #67, opened 2016-08-17
Rya 101 and 102 Mongo Rya example fails to initialize.

DLotts:
Initial fix.  Blocked by RYA-133.  More commits to come.

### DIFF
--- a/dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
+++ b/dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
@@ -21,12 +21,18 @@ package mvm.rya.mongodb;
 
 import java.net.UnknownHostException;
 import java.util.Arrays;
+import java.io.IOException;
 
+import org.apache.commons.configuration.ConfigurationRuntimeException;
 import org.apache.hadoop.conf.Configuration;
 
 import com.mongodb.MongoClient;
 import com.mongodb.MongoCredential;
+import com.mongodb.MongoException;
 import com.mongodb.ServerAddress;
+
+import de.flapdoodle.embed.mongo.distribution.Version;
+import de.flapdoodle.embed.mongo.tests.MongodForTestsFactory;
 
 /**
  * Mongo convention generally allows for a single instance of a {@link MongoClient}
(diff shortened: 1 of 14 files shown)


dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
+                    MongodForTestsFactory testsFactory = MongodForTestsFactory.with(Version.Main.PRODUCTION);
+                    mongoClient = testsFactory.newMongo();
+                } catch (IOException e) {
+                    // Rethrow as an unchecked error.  Since we are in a test mode here, just fail fast.
+                    throw new MongoException("While creating a factory for a test/mock MongoDB instance.",e);
+                }
             } else {
-                mongoClient = new MongoClient(server);
+                // Connect to a running Mongo server
+                final String host = requireNonNull(conf.get(MongoDBRdfConfiguration.MONGO_INSTANCE), "MongoDB host name is required");
```
amihalik: Use guava preconditions instead (e.g. [here](https://github.com/apache/incubator-rya/blob/7743a42a5febe440107e037aadc3753643708e83/common/rya.api/src/main/java/mvm/rya/api/RdfCloudTripleStoreConfiguration.java#L167))

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
-            if (conf.get(MongoDBRdfConfiguration.MONGO_USER) != null) {
-                final String username = conf.get(MongoDBRdfConfiguration.MONGO_USER);
-                final String dbName = conf.get(MongoDBRdfConfiguration.MONGO_DB_NAME);
-                final char[] pswd = conf.get(MongoDBRdfConfiguration.MONGO_USER_PASSWORD).toCharArray();
-                final MongoCredential cred = MongoCredential.createCredential(username, dbName, pswd);
-                mongoClient = new MongoClient(server, Arrays.asList(cred));
+    public static synchronized MongoClient getMongoClient(final Configuration conf)
+            throws ConfigurationRuntimeException, MongoException {
+        if (mongoClient == null) {
+            // The singleton client has not yet created, is it a test/mock instance, or a service?
```
isper3at: this doc is misleading.  There is only one, but it isn't a singleton.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
+            if (conf.getBoolean(MongoDBRdfConfiguration.USE_TEST_MONGO, false)) {
```
isper3at: break these out into their own methods, getTestMongoClient()/getMongoClient()

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
+                } catch (UnknownHostException e) {
+                    throw new MongoException("Connecting MongoDB instance, cannot find host="+host,e);
+                }
+                // check for authentication credentials
+                if (conf.get(MongoDBRdfConfiguration.MONGO_USER) != null) {
+                    final String username = conf.get(MongoDBRdfConfiguration.MONGO_USER);
+                    final String dbName = requireNonNull(conf.get(MongoDBRdfConfiguration.MONGO_DB_NAME),
+                            MongoDBRdfConfiguration.MONGO_DB_NAME + " is null but required configuration if "
+                                    + MongoDBRdfConfiguration.MONGO_USER + " is configured.");
+                    final char[] pswd = requireNonNull(conf.get(MongoDBRdfConfiguration.MONGO_USER_PASSWORD),
```
isper3at: the requireNonNull() should be before this.  assignment should only happen after the precondition checks.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
+
+    /**
+     * Throw exception for un-configured required values.
+     * 
+     * @param required  String to check
+     * @param message  throw configuration exception with this description
+     * @return unaltered required string
+     * @throws ConfigurationRuntimeException  if required is null
+     */
+    static private String requireNonNull(String required, String message) throws ConfigurationRuntimeException {
```
isper3at: what Aaron said.  This is not needed

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -41,26 +47,77 @@
...
+        if (required == null)
+            throw new ConfigurationRuntimeException(message);
+        return required;
+    }
+
+    /*
+     * Same as above, check that it is a integer and return the parsed integer.
+     */
+    static private int requireNonNullInt(String required, String message) throws ConfigurationRuntimeException {
```
isper3at: same as above, as a general note scope is the first keyword, then static.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoDBRyaDAO.java
```diff
@@ -73,14 +73,9 @@
...
         this.conf = conf;
-        try {
-            mongoClient = MongoConnectorFactory.getMongoClient(conf);
-            conf.setMongoClient(mongoClient);
-            init();
-        } catch (NumberFormatException | UnknownHostException e) {
-            log.error("Unable to create a connection to mongo.", e);
-            throw e;
-        }
+        mongoClient = MongoConnectorFactory.getMongoClient(conf);
```
isper3at: why did you remove the log?

extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
```diff
@@ -99,29 +99,19 @@ public void setConf(final Configuration conf) {
         this.conf = conf;
         //reset the init.
         init = false;
-        try {
             init();
-        } catch (final NumberFormatException | UnknownHostException e) {
-            LOG.error("Unable to update to use new config, falling back to the old config.", e);
```
isper3at: same here

extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
```diff
@@ -99,29 +99,19 @@ public void setConf(final Configuration conf) {
...
             if (ConfigUtils.getUseMongo(conf)) {
-                try {
                     final MongoClient mongoClient = MongoConnectorFactory.getMongoClient(conf);
                     geoIndexer = new MongoGeoIndexer(mongoClient);
                     geoIndexer.setConf(conf);
                     freeTextIndexer = new MongoFreeTextIndexer(mongoClient);
                     freeTextIndexer.setConf(conf);
                     temporalIndexer = new MongoTemporalIndexer(mongoClient);
                     temporalIndexer.setConf(conf);
-                } catch (NumberFormatException | UnknownHostException e) {
```
isper3at: same here

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
DLotts: I started with that.  (Objects.requireNonNull() is similar)   But I wanted to throw a different exception, not NullPointerException since the caller may want to specifically catch it.  So then I wrapped the whole thing in a try with a catch that re-throws.  But then it catches ANY null pointer, not just the ones I checked and it became more complicated, so I did it this way.  What do you think, is it worth it to throw a different unchecked error?

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
DLotts: In addition to what I said to Aaron:  I am preserving the difference between a missing configuration and a random NullPointerException and NumberFormatException.  
I'll switch it back if you all consider this unimportant.  The method now throws two unchecked, and IMO not recoverable errors:  ConfigurationRuntimeException for configuration parameters that are missing, and MongoException for issues preventing connection, possibly temporary.  But those issues can be specifically caught, for example: to retry.   A NullPointerException could be thrown for any reason.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoDBRyaDAO.java
DLotts: the UnknownHostException is a checked exception and must be caught.  I removed both of the exceptions and wrapped into better unchecked exceptions.  I assume that the default error handler will log the error.  Otherwise we should attempt to log every unchecked exception.  I made the underlying thrown errors more descriptive, they all mention that the MongoDB connection has failed, so this log entry is redundant.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
isper3at: This seems like an abuse of try/catch.  You generally do a check on the config values way before this point.  Doing it this way can also be super annoying if multiple values are missing.  Run, break, missing a.  Run, break, missing c.  Run, break, missing e, etc.

extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
DLotts: Same reason.  But in addition, this is just a bug and a fib.  It ignores what error happened and any diagnostic information.  It does not restore the old config if there was one. It sets init = false; so that it does not attempt to re-initialize.  The next thing that happens is a failure in distant code that assumes that the database is correctly connected.

extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
DLotts: Same as the first one.  Now redundant, and the catch is no longer required.

dao/mongodb.rya/src/main/java/mvm/rya/mongodb/MongoConnectorFactory.java
DLotts: Private methods called from this method, correct?  Great idea.

DLotts:
This is complete and ready to review.

DLotts:
This was pulled into apache:develop !

PR #68, opened 2016-08-18
[WIP] RYA-151 Rya Query benchmark tool implemented using JMH.

kchilton2:


extras/rya.benchmark/src/test/java/org/apache/rya/benchmark/query/BenchmarkQueriesReaderIT.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class BenchmarkQueriesReaderIT {
+
+    @Test
+    public void load() throws JAXBException, SAXException {
+        // Unmarshal some XML.
+        final String xml =
+                "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
+                "<BenchmarkQueries>\n" +
+                "    <SPARQL><![CDATA[SELECT ?a WHERE { ?a <http://likes> <urn:icecream> . }]]></SPARQL>" +
```
amihalik: Can we add attributes to this?  for instance

```xml
<BenchmarkQueries>
   <SPARQL name="ice cream query" description="This benchmarks the ice cream index"><![CDATA[SELECT ?a WHERE { ?a <http://likes> <urn:icecream> . }]]></SPARQL>
<BenchmarkQueries>

```

kchilton2: What do you want to do with those attributes? If they aren't used within the benchmark, that sort of information could easily just be XML comments.

{code}
<BenchmarkQueries>
    
    
   <SPARQL><![CDATA[SELECT ?a WHERE { ?a <http://likes> <urn:icecream> . }]]></SPARQL>
<BenchmarkQueries>
{code}

amihalik: We should have query level granularity for the benchmark (i.e. we should have timing results for each query in the benchmark), so that we can have pretty graphs like this:

image


Also, it would be nice if the benchmark printed out the name of the query for the results.  So we can have a table like this:

image

(Note: I ripped off these graphics from Roshan's Paper)

kchilton2: The data is reported by the JMH framework for every parameter you tweak. So it will say which query the results are for. You can use the benchmark tool's output to generate those reports.

kchilton2: So the two parameters that I was planning on having you configure are:
1. What SPARQL query to execute.
2. The number of results to read from the query.

amihalik:

PR #70, opened 2016-08-18
[WIP] RYA-109 Add Spark Support

Ejgood21:
Spark GraphX Support - WIP


amihalik:
Something is screwy with Jesse's commit that you pulled in.  The commit you have has hash 	1855213 and the one in the apache/develop branch has 01489ef.  This needs to be fixed before we can pull in your PR, and it's probably easier to fix now rather than later.  Let me know if you need any assistance with this and I can walk you through it.

PR #72, opened 2016-08-18
RYA-133 Fix broken Mongo Secondary Indexers

isper3at:
The secondary indexers need an empty constructor.  Moved back to an init() call instead, leave empty constructor open for the reflexive calls.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-133)

### Checklist
- [ ] Code Review
- [x] Squash commits

#### People To Review
@dlotts

### DIFF
--- a/extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
+++ b/extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
@@ -68,6 +68,7 @@ import mvm.rya.indexing.accumulo.freetext.FreeTextTupleSet;
 import mvm.rya.indexing.accumulo.geo.GeoMesaGeoIndexer;
 import mvm.rya.indexing.accumulo.geo.GeoTupleSet;
 import mvm.rya.indexing.accumulo.temporal.AccumuloTemporalIndexer;
+import mvm.rya.indexing.mongodb.AbstractMongoIndexer;
 import mvm.rya.indexing.mongodb.freetext.MongoFreeTextIndexer;
 import mvm.rya.indexing.mongodb.geo.MongoGeoIndexer;
 import mvm.rya.indexing.mongodb.temporal.MongoTemporalIndexer;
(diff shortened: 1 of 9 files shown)


extras/indexing/src/main/java/mvm/rya/indexing/FilterFunctionOptimizer.java
```diff
@@ -101,23 +102,23 @@ public void setConf(final Configuration conf) {
...
                 try {
                     final MongoClient mongoClient = MongoConnectorFactory.getMongoClient(conf);
-                    geoIndexer = new MongoGeoIndexer(mongoClient);
-                    geoIndexer.setConf(conf);
-                    freeTextIndexer = new MongoFreeTextIndexer(mongoClient);
-                    freeTextIndexer.setConf(conf);
-                    temporalIndexer = new MongoTemporalIndexer(mongoClient);
-                    temporalIndexer.setConf(conf);
+                    geoIndexer = new MongoGeoIndexer();
+                    ((AbstractMongoIndexer) geoIndexer).initIndexer(conf, mongoClient);
```
DLotts: GeoIndexer has null fields: storageStrategy , predicates.  The implementation init is not called?  In the code I see from develop, init does this 
<code>predicates = ConfigUtils.getGeoPredicates(conf);  
storageStrategy = ...<code>

isper3at: that might be a problem in the other initIndexer()  The reason there is a seperate initIndexer() function is because init() does things I don't want to happen quite yet.

amihalik:
No comments on your PR commit, but I really like your description.  I think we should do this: [RYA-156 - Add a Pull Request Template](https://issues.apache.org/jira/browse/RYA-156)

isper3at:
I already have it templated up in a bookmarklett.  We can actually add something to the .git folder so that the template is auto-added to the PR.  I'll assign the ticket to myself and set it up

amihalik:
Merged.

PR #77, opened 2016-08-22
RYA-153 Fixed JoinPCJMatcher bug

meiercaleb:
Validate input to JoinSegmentPCJMatcher to prevent unpredictable behavior when it attempts to match queries with LeftJoins.


amihalik:

amihalik:

PR #78, opened 2016-08-22
RYA-154 Decoupled Updater Configuration from PCJ Configuration

meiercaleb:
Decoupled PrecomputedJoinIndexer configuration from PCJ configuration.  Users who want to use PCJs (which is a query based configuration) should not also be using the PrecomputedJoinIndexer (which affects ingest).


amihalik:

PR #80, opened 2016-08-24
RYA-163 PCJDetails Test Failures on non-EST TimeZones

DLotts:
This bug was found by jenkins.

### DIFF
--- a/extras/rya.console/src/test/java/mvm/rya/shell/RyaAdminCommandsTest.java
+++ b/extras/rya.console/src/test/java/mvm/rya/shell/RyaAdminCommandsTest.java
@@ -27,6 +27,7 @@ import static org.mockito.Mockito.when;
 import java.io.IOException;
 import java.util.Date;
 import java.util.List;
+import java.util.TimeZone;
 
 import org.junit.Test;
 
@@ -123,6 +124,9 @@ public class RyaAdminCommandsTest {
 
     @Test
     public void getInstanceDetails() throws InstanceDoesNotExistException, RyaClientException {
+        // This test is failed if the default timezone was not EST, so now it's fixed at EST.
+        // If you get assert mismatch of EST!=EDT, try the deprecated getTimeZone("EST") instead.
+        TimeZone.setDefault(TimeZone.getTimeZone("America/New_York"));
         // Mock the object that performs the get operation.
         final GetInstanceDetails mockGetInstanceDetails = mock(GetInstanceDetails.class);
         final String instanceName = "test_instance";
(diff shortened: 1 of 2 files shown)


DLotts:
This is complete and ready for review and pull.

amihalik:
Nice... super simple.  I tested it out and it passed on my machine.

DLotts:
Done again.  Second failed test is fixed using the same method.

extras/rya.console/src/test/java/mvm/rya/shell/RyaAdminCommandsTest.java
```diff
@@ -123,6 +124,9 @@ public void deletePCJ() throws InstanceDoesNotExistException, RyaClientException
 
     @Test
     public void getInstanceDetails() throws InstanceDoesNotExistException, RyaClientException {
+        // This test is failed if the default timezone was not EST, so now it's fixed at EST.
+        // If you get assert mismatch of EST!=EDT, try the deprecated getTimeZone("EST") instead.
+        TimeZone.setDefault(TimeZone.getTimeZone("America/New_York"));
```
amihalik: You made a good point in in the JIRA ticket... I think everything worked for me because the order of the test execution.  Can you move this to a static class initializer or a BeforeClass method (so that the order of the test within this file doesn't break things)?

DLotts: These are two different classes.  Before will still have an order dependence.  
Also every test after will still be affected.
Another solution is to replace it to the original value when the test is over.
With that, only concurrent unit tests will have order dependence, in that case, a race condition.  Then we need to synchronize on the timezone setting.
But what is wrong with setting it everywhere there is a dependence on TimeZone?  All the issues with that all have to do with setting the timezone default globally, but not order, nor concurrency -- unless someone sets it to a non EST timezone, then the concurrency issue comes back, but not the serial order.  

So I think it is best as it is now -- not perfect, but simple.  What do you think?

kchilton2: Alternatively, couldn't commands.getInstanceDetails(); accept the time zone as a parameter?

amihalik: I think there is a bigger issue... the setDetails() methods should be using java 8 Time libraries.  I haven't used those yet, but I'm guessing we want to be using java.time.Instant instead java.util.Date

DLotts: Yes completely agree.  Java 8 or JodaTime.  They are mostly the same.  I would only use  java.util.Date if I had to.

PR #85, opened 2016-09-06
RYA-151 - Benchmarking Tools

kchilton2:


extras/indexing/src/main/java/mvm/rya/indexing/external/PrecomputedJoinStorageSupplier.java
```diff
@@ -27,6 +27,7 @@
 import com.google.common.base.Optional;
 import com.google.common.base.Supplier;
 
+import cern.colt.Arrays;
```
amihalik: Is this correct?  Or do you mean `java.util.Arrays`

extras/rya.benchmark/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/xsd</directory>
+            </resource>
+        </resources>
+    
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
amihalik: this should be define in the project's root pom

extras/rya.benchmark/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+    <artifactId>rya.benchmark</artifactId>
+
+    <name>Apache Rya Benchmarks</name>
+
+    <dependencies>
+        <!-- JMH Benchmark Framework dependencies -->
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>${jmh.version}</version>
```
amihalik: These versions should be defined in the root pom as properties and in the root pom's plug-in management section

extras/indexing/src/main/java/mvm/rya/indexing/external/PrecomputedJoinStorageSupplier.java
kchilton2: That's likely a mistake. Good catch.

extras/rya.benchmark/pom.xml
kchilton2: Which part specifically?

amihalik: Sorry for my terse comment.  These should be defined in the parent POM:

1. Versions for dependencies should be defined as parameters (e.g. like you've done with jmh.version).  You should also do this for maven plug-ins (e.g. you haven't done this for maven-clean-plugin).

2. Dependencies (and plug-ins) and their versions should be defined in the DependenciesManagement (and pluginManagement) sections of the root pom.

3. The root pom should be the only pom to define the java versions.  Subprojects shouldn't be using the maven-compiler-plugin to define their own versions of java.

4. You have some nice plug-ins that should be moved to the root pom as well (e.g. license-maven-plugin).  I think @DLotts is working on this, so there's no need for you to modify that here.

kchilton2: This POM file was largely generated by as a JMH archetype and I'm not sure what I can safely change without breaking how the JMH stuff works.

kchilton2: Fixed what I know can be moved. It's in the new commit.

extras/indexing/src/main/java/mvm/rya/indexing/external/PrecomputedJoinStorageSupplier.java
kchilton2: Fixed, new commit.

PR #118, opened 2016-10-21
RYA-174 Removing Mongo Test Factory from non-test code

amihalik:
> What Changed?

Removed Mongo Test Factory code from non-test code
### Tests

no new tests
### Links

[Jira](https://issues.apache.org/jira/browse/RYA-174)
### Checklist
- [ ] Code Review
- [ ] Squash Commits
#### People To Reivew

@dlotts

### DIFF
--- a/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MockMongoFactory.java
+++ b/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MockMongoFactory.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.rya.mongodb;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.net.UnknownHostException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.mongodb.MongoClient;
+import com.mongodb.MongoException;
+import com.mongodb.ServerAddress;
+
+import de.flapdoodle.embed.mongo.Command;
+import de.flapdoodle.embed.mongo.MongodExecutable;
+import de.flapdoodle.embed.mongo.MongodProcess;
+import de.flapdoodle.embed.mongo.MongodStarter;
+import de.flapdoodle.embed.mongo.config.IMongodConfig;
+import de.flapdoodle.embed.mongo.config.MongodConfigBuilder;
+import de.flapdoodle.embed.mongo.config.Net;
+import de.flapdoodle.embed.mongo.config.RuntimeConfigBuilder;
+import de.flapdoodle.embed.mongo.distribution.IFeatureAwareVersion;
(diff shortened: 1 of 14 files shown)


isper3at:
I disagree with the notion of matching accumulo naming conventions. I'm fine with ambiguous naming in interfaces and during design, but altering naming to match another datastore just seems.....wrong?  If we want consistency, lets use what flapdoodle calls it.  Then we already have the correct naming if we need to google something.

pujav65:
hey i see where this was coming from, but i think we should try to move to the mongo client being passed in.  I'd like to see this look a lot closer to what was done in Accumulo (more because the Accumulo implementation is in line with what the intended contracts of the RyaDAO/etc. are).   We started moving away from that when we started having a connections factory -- I think this pull request removes that which is good, but I agree with Aaron that we should move to have the connector/client be passed in to everything that needs it rather than passing in the connection information.

isper3at:
what do you mean by 'removes that which is good'?  The connections factory is a placeholder until we can get rid of the dependency on having the hadoop configuration everywhere (i put it in the docs the first time it went in).  Accumulo has had a factory style object for a while now, the AccumuloInstanceDriver is essentially a factory

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MockMongoFactory.java
```diff
@@ -0,0 +1,97 @@
...
+import de.flapdoodle.embed.mongo.config.MongodConfigBuilder;
+import de.flapdoodle.embed.mongo.config.Net;
+import de.flapdoodle.embed.mongo.config.RuntimeConfigBuilder;
+import de.flapdoodle.embed.mongo.distribution.IFeatureAwareVersion;
+import de.flapdoodle.embed.mongo.distribution.Version;
+
+public class MockMongoFactory {
+    private static Logger logger = LoggerFactory.getLogger(MockMongoFactory.class.getName());
+
+    public static MockMongoFactory newFactory() throws IOException {
```
amihalik: hrm.  I couldn't find that class.  I won't consolidate the classes in this PR.  Let's do it in another Jira Issue.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MockMongoFactory.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    /**
+     * Creates a new Mongo connection.
+     * 
+     * @throws MongoException
+     * @throws UnknownHostException
+     */
+    public MongoClient newMongo() throws UnknownHostException, MongoException {
```
amihalik: Done.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MongoConnectorFactory.java
```diff
@@ -56,28 +52,14 @@
...
-     */
-    private static void createMongoClientForTests() throws MongoException {
-        try {
-            MongodForTestsFactory testsFactory = MongodForTestsFactory.with(Version.Main.PRODUCTION);
-            mongoClient = testsFactory.newMongo();
-        } catch (IOException e) {
-            // Rethrow as an unchecked error.  Since we are in a test mode here, just fail fast.
-            throw new MongoException(MSG_INTRO+"creating a factory for a test/mock MongoDB instance.",e);
-        }
+    public static synchronized void closeMongoClient() {
```
amihalik: Done.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/MockMongoFactory.java
isper3at: Oh, its not in yet.  Its in my RYA-123 pull request

amihalik:
It seems like everyone is on board with this PR... I'm going to let Jenkins chomp on it for a bit and (unless I hear otherwise) pull this into master after the build.

pujav65:
i thought the mongo connections factory was a bad idea at the time.  so i think removing it is a good thing in general.  The instance driver you're referring to is in the merge tool -- its not used in the main client path.  The accumulo dao has a setConnector() method which really should have been emulated by the Mongo dao.   I didn't realize the pattern until we started having issues with setting up multiple connections, but the passing in the connector/client pattern is nice because its a lot more flexible than what we currently have.  Regardless, I think this pull request is fine -- all of this discussion is on future direction.

PR #122, opened 2016-11-21
adding metadata to individual statements

pujav65:
[Brief Description of what changed]

[Description of what tests were written]

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-NUMBER)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew

### DIFF
--- a/common/rya.api/src/main/java/org/apache/rya/api/domain/StatementMetadata.java
+++ b/common/rya.api/src/main/java/org/apache/rya/api/domain/StatementMetadata.java
@@ -1,0 +1,73 @@
+package org.apache.rya.api.domain;
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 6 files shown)


common/rya.api/src/main/java/org/apache/rya/api/domain/StatementMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+    private static Gson gson = new Gson();
+    public static StatementMetadata EMPTY_METADATA = new StatementMetadata();
+    
+    private Map<String, String> metadataMap = new HashMap<String, String>();
+    
+    public StatementMetadata() {
+        
+    }
+    
+    public void addMetadata(String key, String value){
```
amihalik: This behaves like a "put" and not an "add".  Consider renaming this method.

PR #130, opened 2016-12-17
corrected errors in sample code in README.md

johns0806:
## Description
Corrected the three code snippets to be easier for new users to use.  Fixed broken sample code that used non-existent methods.  Made "rya_" prefix consistent between code snippets.  Greatly simplified the sample code which writes to Rya by removing many lines of unneeded code while retaining the same capability.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-229)


amihalik:
asfbot build

pujav65:
asfbot build

PR #135, opened 2016-12-30
Rya 144

isper3at:
-Finished kchilton2's mongo based entity secondary indexer
-updated to use meiercaleb's new query segment framework
-added optimizer for the entity indexer

Added integration tests for the indexer
Note: no assertions are performed

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-144)

### Checklist
- [x] Code Review
- [ ] Squash Commits

#### People To Reivew
@meiercaleb 
@pujav65


extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexSetProvider.java
```diff
@@ -0,0 +1,136 @@
...
+                //if is type, fetch type add to subject -> type map
+                if(optType.isPresent()) {
+                    final Type type = optType.get();
+                    typeMap.put(type, pattern);
+                    subjectTypeMap.put(subjStr, type);
+                    //check unmatched properties, add matches
+                    for(final StatementPattern propertyPattern : unmatched) {
+                        //store sp's as uri?
+                        final RyaURI property = getPredURI(propertyPattern);
+                        final Var typeSubVar = getTypeSubject(type);
```
meiercaleb: Why call getTypeSubject here to extract the subject Var from typeMap?  You already have access to subjStr.  Why not just call propertyPattern.getSubjectVar().getName() below and do a String comparison in your if clause?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexSetProvider.java
```diff
@@ -0,0 +1,136 @@
...
+            //if not type, check to see if subject is in type map
+            if(subjectTypeMap.containsKey(subjStr)) {
+                //if is, check to see if pred is a property of type
+                final Type type = subjectTypeMap.get(subjStr);
+                if(type.getPropertyNames().contains(predURI)) {
+                    //if is, add sp to type -> sp map
+                    typeMap.put(type, pattern);
+                } else {
+                    //if not, add to unmatched type
+                    unmatched.add(pattern);
```
meiercaleb: I'm not so sure that this is what you want to do with a pattern whose subject name corresponds to a given type, but whose property doesn't match any of the type properties.  By adding it to unmatched you are allowing for the possibility that this pattern can be matched with another type corresponding to the same variable.  It seems as though you don't want to allow this given that subjectTypeMap is a hashmap (so you can associate only one type to each subject var name).  On the other hand, if you have are fed a segment that has two types associated with the same variable then your entry in subjectTypeMap will be overwritten and it is possible for you to have StatementPatterns assigned to each of the types in typeMap.  So I would add some sort of check to ensure you don't have multiple types associated with the same subject variable in your segment.  Once this is in place you don't need to add the above pattern to unmatched because there is no chance it will be matched to another type with the same variable.  With that in mind, you could probably revise the above branching structure to be if( ... equals(RDF.TYPE)) {...} else if(subjectTypeMap.containsKey(subjStr)) {...} else {unmatched.add(pattern)}

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityToSegmentConverter.java
```diff
@@ -0,0 +1,109 @@
...
+        //}
+    }
+
+    /**
+     * This class extracts the {@link JoinSegment} from the {@link TupleExpr} of
+     * specified PCJ.
+     *
+     */
+    static class EntityToJoinSegment extends QueryModelVisitorBase<RuntimeException> {
+
```
meiercaleb: There is no tree structure to your EntityQueryNode.  Why are you using a visitor here?  Probably just a copy and paste issue.  Migrate the code in lines 57 - 59 into the method starting at 36.   Or just have a getSegment method.  This code does not need to be contained inside two distinct visitor classes (you're not even using one of them!)

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityToSegmentConverter.java
```diff
@@ -0,0 +1,109 @@
...
+            segment = new JoinSegment<EntityQueryNode>(filter);
+        }
+
+    }
+
+    /**
+     * This class extracts the {@link OptionalJoinSegment} of PCJ query.
+     *
+     */
+    static class EntityToOptionalJoinSegment extends QueryModelVisitorBase<RuntimeException> {
```
meiercaleb: Get rid of this.  See above comment.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexOptimizer.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final Logger log = Logger.getLogger(EntityIndexOptimizer.class);
+    private static final EntityExternalSetMatcherFactory MATCHER_FACTORY = new EntityExternalSetMatcherFactory();
+    private final EntityIndexSetProvider provider;
+
+    /**
+     * Creates a new {@link EntityIndexOptimizer}.
+     * @param typeStorage - The mechanism for access the various {@link Type}s. (not null)
+     * @param entityStorage - The mechanism for access to the various {@link Entity}s. (not null)
+     */
+    public EntityIndexOptimizer(final TypeStorage typeStorage, final EntityStorage entityStorage) {
```
meiercaleb: Unfortunately, you need to be able to create the Optimizer from  a Configuration object.  See RdfCloudTripleStoreConnection 316-335.  This is where your optimizer will be used.  You also need to add setters and getters to the MongoIndexingConfiguration object to configure Rya to add your optimizer as an additional optimizer.

isper3at: freakin kchilton2.....

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityToSegmentConverter.java
isper3at: This may be an artifact of what I did before and after your refactor.  if its isolated not used code maven is happy with it.  same below

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexSetProvider.java
isper3at: good catch.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexSetProvider.java
isper3at: I think because String comparison is heavy and I'm hoping the Var comparison isn't.

isper3at:
Looks likeEntityIndexOptimizer needs fixing.  And I need to add the optimizer to some config stuff.  I'll get on that and update the PR

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -0,0 +1,356 @@
...
+                throw new IllegalStateException("The Predicate of a Statement Pattern must be a property name for the Type. " +
+                        "Type ID: '" + type.getId().getData() + "' Pattern: " + pattern);
+            }
+        }
+    }
+
+    @Override
+    public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(final Collection<BindingSet> bindingSets) throws QueryEvaluationException {
+        // TODO Auto-generated method stub
+        return null;
```
pujav65: Is there a reason this isn't implemented?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/storage/CloseableIterator.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+/**
+ * An {@link Iterator} that is also {@link Closeable}.
+ *
+ * @param <T> - The type of object that will be iterated over.
+ */
+public interface CloseableIterator<T> extends Iterator<T>, Closeable { }
```
pujav65: why did we re-implement mango's closeableiterator
(org.calrissian.mango.collect.closeableiterator)?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/update/EntityIndexer.java
```diff
@@ -0,0 +1,30 @@
...
+package org.apache.rya.indexing.entity.update;
+
+import org.apache.rya.api.persist.index.RyaSecondaryIndexer;
+import org.apache.rya.indexing.entity.storage.EntityStorage;
+
+/**
+ * Updates the {@link Entity}s that are in a {@link EntityStorage} when new
+ * {@link RyaStatement}s are added/removed from the Rya instance.
+ */
+public interface EntityIndexer extends RyaSecondaryIndexer {
```
pujav65: why doesn't this have any methods?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: I didn't see it getting called from anywhere, its an artifact of what kchilton2 did.  No idea what would go in here

extras/indexing/src/main/java/org/apache/rya/indexing/entity/storage/CloseableIterator.java
isper3at: not sure, I think this was brought in in meiercaleb's commit

extras/indexing/src/main/java/org/apache/rya/indexing/entity/update/EntityIndexer.java
isper3at: Again, not mine.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/storage/CloseableIterator.java
isper3at: this was kchilton2's if anything I think its because we only want Closeable and Iterator, none of the other functionality, if any, from mango's.  I don't think this is supposed to be a re-implementation of mango's.  I think its just the same name

pujav65: ok, can you please remove it and then refactor places its used to use the mango version?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/update/EntityIndexer.java
pujav65: i would just remove this then -- either that or pull up some methods that indicate its intent, i'm not really sure what having an empty interface buys us since its not even particularly well documented.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/storage/CloseableIterator.java
isper3at: I think the mango version has functionality we don't necessarily want

extras/indexing/src/main/java/org/apache/rya/indexing/entity/update/EntityIndexer.java
isper3at: moved up the get storage functions.  the interface makes it way easier to test/mock.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/storage/CloseableIterator.java
pujav65: i'm not good with merging this until this is either removed or renamed to not be the same name as the mango implementation.  it reminds me of places where you see custom lists (like arraylist), which i would like to avoid introducing if we can.
If you won't refactor to use the mango implementation, then rename to something other than CloseableIterator.  Once that's done I think this is good to merge.

pujav65:
wanted to check if this was ready to merge -- looks like everything has been addressed?

isper3at:
Still need to do the change to CloseableIterable

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -20,31 +20,47 @@
...
 
+import edu.umd.cs.findbugs.annotations.DefaultAnnotation;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import info.aduna.iteration.CloseableIteration;
-import mvm.rya.api.domain.RyaURI;
-import mvm.rya.rdftriplestore.evaluation.ExternalBatchingIterator;
+import info.aduna.iteration.EmptyIteration;
 
 /**
  * TODO impl, test, doc
```
kchilton2: Missing docs.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -62,11 +78,14 @@
     private final Optional<String> subjectConstant;
     private final Optional<String> subjectVar;
 
+    //since and EntityQueryNode exists in a single segment, all binding names are garunteed to be assured.
+    private final Set<String> bindingNames;
+
     // Information about the objects of the patterns.
 
     // XXX what does this map? property name -> binding variable?
```
kchilton2: Cleanup docs.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -105,8 +126,24 @@ public EntityQueryNode(final Type type, final Collection<StatementPattern> patte
         // TODO Also, map each variable that is in an Object spot each variable can be mapped to a property name as well
```
kchilton2: Has this TODO been addressed?

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -204,35 +241,116 @@ private static void verifyAllPredicatesPartOfType(Type type, Collection<Statemen
...
         // TODO Auto-generated method stub
         return null;
     }
 
     @Override
-    public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(BindingSet bindingSet) throws QueryEvaluationException {
+    public CloseableIteration<BindingSet, QueryEvaluationException> evaluate(final BindingSet bindingSet) throws QueryEvaluationException {
         requireNonNull(bindingSet);
-
         // ... ok, so if the subject needs to be filled in, then we need to see if the subject variable is in the binding set.
```
kchilton2: Cleanup docs.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -204,35 +241,116 @@ private static void verifyAllPredicatesPartOfType(Type type, Collection<Statemen
...
+                    }
+                }
+            } else {
+                return new EmptyIteration<BindingSet, QueryEvaluationException>();
+            }
+        } catch (final EntityStorageException e) {
+            throw new QueryEvaluationException("Failed to evaluate the binding set", e);
+        }
+        // x = alice's SSN
+        // y = blue  <-- how do i know this is for urn:eye property? FROM THE STATEMENT PATTERN. look for the ?y in the SP, and that has the property name in it.
```
kchilton2: Cleanup docs.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -204,35 +241,116 @@ private static void verifyAllPredicatesPartOfType(Type type, Collection<Statemen
...
+        }
+        return hashcode;
+    }
 
-        // x = alice's SSN
-        // y = blue  <-- how do i know this is for urn:eye property? FROM THE STATEMENT PATTERN. look for the ?y in the SP, and that has the property name in it.
+    @Override
+    public boolean equals(final Object other) {
+        if(other instanceof EntityQueryNode) {
+            return hashCode() == other.hashCode();
```
kchilton2: This isn't a valid implementation of equals. You can have non-equal things have the same hash via a hash collision.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -204,35 +241,116 @@ private static void verifyAllPredicatesPartOfType(Type type, Collection<Statemen
...
+    /**
+     * @return - The {@link StatementPattern}s that make up this {@link EntityQueryNode}
+     */
+    public Collection<StatementPattern> getPatterns() {
+        return patterns;
+    }
 
-        // for all variables in the OBJECT portion of the SPs, fill 'em in using the entity that is stored in the index.
+    @Override
+    public int hashCode() {
```
kchilton2: Maybe use the Objects.hash helper function here instead of building your own.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
```diff
@@ -204,35 +241,116 @@ private static void verifyAllPredicatesPartOfType(Type type, Collection<Statemen
...
+                            final Value objectVal = object.getValue();
+                            if(objectVal.equals(entityVal)) {
+                                //add to binding set
+                                resultSet.addBinding(object.getName(), entityVal);
+                            }
+                        } else {
+                            resultSet.addBinding(object.getName(), entityVal);
+                        }
+                    } else {
+                        //entity doesn't hold to statement patterns
```
kchilton2: No point in having an empty else.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexOptimizer.java
```diff
@@ -0,0 +1,128 @@
...
+    private final MongoEntityIndexer indexer;
+    private EntityIndexSetProvider provider;
+    private Configuration conf;
+
+    private EntityStorage entityStorage;
+    private TypeStorage typeStorage;
+
+    /**
+     * Creates a new {@link EntityIndexOptimizer}.
+     * @param typeStorage - The mechanism for access the various {@link Type}s. (not null)
```
kchilton2: Your docs do not match the constructor signature.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexOptimizer.java
```diff
@@ -0,0 +1,128 @@
...
+        checkNotNull(indexer);
+
+        // first standardize query by pulling all filters to top of query if
+        // they exist using TopOfQueryFilterRelocator
+        tupleExpr = TopOfQueryFilterRelocator.moveFiltersToTop(tupleExpr);
+        super.optimize(tupleExpr, null, null);
+    }
+
+    @VisibleForTesting
+    public EntityStorage getEntityStorage() {
```
kchilton2: This methods that only exist for testing are a smell. You should be able to push a mock or the object you need access to within your tests into this class.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexSetProvider.java
```diff
@@ -0,0 +1,157 @@
...
+            try {
+                final Optional<Type> optType = typeStorage.get(objURI);
+                //if is type, fetch type add to subject -> type map
+                if(optType.isPresent()) {
+                    final Type type = optType.get();
+                    typeMap.put(type, pattern);
+                    subjectTypeMap.put(subjStr, type);
+                    //check unmatched properties, add matches
+                    for(final StatementPattern propertyPattern : unmatched) {
+                        //store sp's as uri?
```
kchilton2: Cleanup comment.

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: done

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: done

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: yes, removed it

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: done

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: removed

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: doc no longer applies, removed

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: done

extras/indexing/src/main/java/org/apache/rya/indexing/entity/query/EntityQueryNode.java
isper3at: fixed

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexOptimizer.java
isper3at: fixed

extras/indexing/src/main/java/org/apache/rya/indexing/entity/EntityIndexOptimizer.java
isper3at: left here per our talk

pujav65:
asfbot build

pujav65:
hey @isper3at can you look at what the failure is here and fix it?  I will merge after that.

isper3at:
for some reason the new test didn't show up on my rat report.  Had to flip through all the new files to find it >..< its up and should be good now

PR #136, opened 2017-01-04
RYA-83-select-all-spo-query-support added capability to perform  SELE…

johns0806:
…CT * { s? p? o? }  queries.

## Description

Added NullRowTriplePatternStrategy.java to support queries in which the ?s ?p ?o are all null (unbound)
Also added this Strategy to RyaTripleContext default triple pattern strategies, so that it is used.

### Tests
Performed an integration test, before the patch I wasn't able to perform SELECT * { s? p? o? }  queries, after the patch I can.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-83)


pujav65:
Looks good.  Can you either add a test or add a Jira ticket to add a test for the new class?  Once you've done that, I'll be ok with merging (as long as no one cares that we now allow full table scans).

johns0806:
Tests added.

PR #142, opened 2017-03-03
RYA-253 Converted the rya.prospector class from Groovy to Java. Also added documentation to the project.

kchilton2:


pujav65:
asfbot build

PR #145, opened 2017-03-13
Fix rat issue

isper3at:
Moved the rat exclude up into the pcj fluo parent pom
so Function gets ignored regardless of profile.

Moved rat exclude up one pom.

Ran build locally with geo profile and without

### Links
N/A

### Checklist
- [ ] Code Review
- [x] Squash Commits

#### People To Reivew
@pujav65 
@amihalik

### DIFF
--- a/extras/rya.pcj.fluo/pom.xml
+++ b/extras/rya.pcj.fluo/pom.xml
@@ -74,6 +74,24 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+		<groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+            </plugin>
         </plugins>
+        <pluginManagement>
+            <plugins>
+                <plugin>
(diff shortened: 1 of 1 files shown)


extras/rya.pcj.fluo/pom.xml
```diff
@@ -74,6 +74,16 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <!--  Trivial listing of classes to be loaded via SPI -->
+                        <exclude>rya.pcj.functions.geo/src/main/resources/META-INF/services/org.openrdf.query.algebra.evaluation.function.Function</exclude>
```
amihalik: What about just `<exclude>**/resources/META-INF/services/**</exclude>` so that it isn't tied to just geo?

(note: I'm fine with it as is, too)

pujav65:
asfbot build

PR #147, opened 2017-03-16
RYA-262 Fixed a bug where AccumuloDocIdIndexer throws exceptions when…

kchilton2:
… the configured authorizations are either an empty string or not set.


pujav65:
merged

PR #155, opened 2017-04-15
Resolves Rya 271

logycon:
What Changed?
Addresses https://issues.apache.org/jira/browse/RYA-271
Modified IndexingFunctionRegistry.getResultVarFromFunctionCall to allow for functions of any FUNCTION_TYPE, as follows:

FUNCTION_TYPE type = SEARCH_FUNCTIONS.get(function);
if (type == null)
{ return findBinaryResultVar(args); }
else {
...

Tests

org.apache.rya.indexing.accumulo.customfunction
AccumuloCustomSparqlFunctionTest.java - unit test
CustomSparqlFunction.java - custom function implementation

Custom function registration in:
src/test/resources/META-INF/services/org.openrdf.query.algebra.evaluation.function.Function

Created a custom function http://example.org#mycustomfucnction.
Using following SPARQL to test:

PREFIX fn: http://example.org#
select (fn:mycustomfucnction() as ?res1) (str(fn:mycustomfucnction('Rya')) as ?res2)
where {}

Links

Jira

Checklist

 Code Review
 Squash Commits


amihalik:
ok to test

extras/indexing/src/main/java/org/apache/rya/indexing/IndexingFunctionRegistry.java
```diff
@@ -73,18 +73,22 @@
...
-        case FREETEXT:
-            return findLiteralResultVar(args);
-        case TEMPORAL:
-            return findBinaryResultVar(args);
-        default:
-            return null;
+        if (type == null) {
+        	return findBinaryResultVar(args);
+        }
+        else {
```
kchilton2: This is kind of a style nit.

Since the conditional returns a value, you really don't need the else clause.

extras/indexing/src/main/java/org/apache/rya/indexing/IndexingFunctionRegistry.java
```diff
@@ -73,18 +73,22 @@
...
+	        switch(type) {
+	        case GEO: 
+	            return findBinaryResultVar(args);
+	        case FREETEXT:
+	            return findLiteralResultVar(args);
+	        case TEMPORAL:
+	            return findBinaryResultVar(args);
+	        default:
+	            //return null;
```
kchilton2: Please remove commented out code.

extras/indexing/src/test/java/org/apache/rya/indexing/accumulo/customfunction/AccumuloCustomSparqlFunctionTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.openrdf.repository.Repository;
+import org.openrdf.repository.sail.SailRepository;
+import org.openrdf.sail.Sail;
+
+import org.junit.*;
+
+import org.openrdf.repository.RepositoryConnection;
+import org.openrdf.repository.RepositoryException;
+
+public class AccumuloCustomSparqlFunctionTest {
```
kchilton2: Style nit: Please replace the tab characters with 4 space characters.

extras/indexing/src/test/java/org/apache/rya/indexing/accumulo/customfunction/AccumuloCustomSparqlFunctionTest.java
```diff
@@ -0,0 +1,96 @@
...
+        conf.setTablePrefix("triplestore_");
+        
+        this.sail = RyaSailFactory.getInstance(conf);
+        this.repository = new SailRepository(this.sail);
+        this.conn = this.repository.getConnection();
+        
+    }
+	
+	@Test
+	public void customFunction() {
```
kchilton2: I think you could avoid using ServiceLoader to add your test function by using the following line:
FunctionRegistry.getInstance().add(fooFunction);

Then you could just define the function you're testing with within the test itself instead of having it loaded into any Sail objects that are instantiated within this project's tests.

extras/indexing/src/test/java/org/apache/rya/indexing/accumulo/customfunction/AccumuloCustomSparqlFunctionTest.java
```diff
@@ -0,0 +1,96 @@
...
+		try {
+			TupleQuery query = this.conn.prepareTupleQuery(QueryLanguage.SPARQL, sql);
+		
+			TupleQueryResult qres = query.evaluate();
+			ArrayList<HashMap<String, Value>> reslist = new ArrayList<HashMap<String, Value>>();
+			while (qres.hasNext()) {
+				BindingSet b = qres.next();
+				Set<String> names = b.getBindingNames();
+				HashMap<String, Value> hm = new HashMap<String, Value>();
+				for (Object n : names) {
```
kchilton2: Why loop over objects when the Set's generic type is already knows to be String?

extras/indexing/src/test/java/org/apache/rya/indexing/accumulo/customfunction/AccumuloCustomSparqlFunctionTest.java
```diff
@@ -0,0 +1,96 @@
...
+		String sql = "" 
+				+ "PREFIX fn: <http://example.org#> "
+				+ "select (fn:mycustomfucnction() as ?res1) (str(fn:mycustomfucnction('Rya')) as ?res2) "
+				+ "where {}";
+		try {
+			TupleQuery query = this.conn.prepareTupleQuery(QueryLanguage.SPARQL, sql);
+		
+			TupleQueryResult qres = query.evaluate();
+			ArrayList<HashMap<String, Value>> reslist = new ArrayList<HashMap<String, Value>>();
+			while (qres.hasNext()) {
```
kchilton2: I think this  whole section of the test could be simplified to just ensuring a BindingSet with the two expected bindings was the result of the query you issue on line 59.

pujav65:
looks like all of the comments were addressed -- I'm good to merge this.

pujav65:
asfbot build

extras/indexing/src/test/java/org/apache/rya/indexing/accumulo/customfunction/AccumuloCustomSparqlFunctionTest.java
```diff
@@ -0,0 +1,81 @@
...
+        conf.setClass(ConfigUtils.TOKENIZER_CLASS, SimpleTokenizer.class, Tokenizer.class);
+        conf.setTablePrefix("triplestore_");
+
+        this.sail = RyaSailFactory.getInstance(conf);
+        this.repository = new SailRepository(this.sail);
+        this.conn = this.repository.getConnection();
+    }
+
+    @Test
+    public void customFunction() {
```
meiercaleb: This test passes without incorporating the changes to the IndexingFunctionRegistry above, so there is no need to merge the changes.  I think that you have have confused the IndexingFunctionRegistry with the FunctionRegistry.
The IndexingFunctionRegistry handles a very specific collection of "Filter Functions" that Rya uses for specific types of Indexing.   The reason I put "Filter Functions" in quotes is that an optimizer keys off of specific URIs that are passed in via SPARQL Filters.  These URIs are affiliated with specific indexes, and the IndexingFunctionRegistry is used to inform the optimizer which index to use based on the URI that is sees.  This particular subset of Functions that is tracked by the IndexingFunctionRegistry is not evaluated like normal RDF4J Functions.
To register custom Functions with RDF4J (and Rya), you simply create the Function and register it with the FunctionRegistry, which you did in your test below.  That is all that is required, and your test passes without the above changes to the IndexingFunctionRegistry.

PR #156, opened 2017-04-18
RYA-260 Fluo SPARQL Processing Aggregation Support

kchilton2:
Added Aggregation support to the Fluo PCJ application. It is a known problem that the Accumulo PCJ Index does not properly handle aggregation output exporting. That should be covered by a new ticket. Also fixed a bunch of resource leaks and changed up how the Fluo integration tests were implemented so that they use the Fluo recipe for Accumulo exporting integration tests.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-260)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew
@meiercaleb


dao/accumulo.rya/src/main/java/org/apache/rya/accumulo/instance/AccumuloRyaInstanceDetailsRepository.java
```diff
@@ -89,12 +89,17 @@ public AccumuloRyaInstanceDetailsRepository(final Connector connector, final Str
 
     @Override
     public boolean isInitialized() throws RyaDetailsRepositoryException {
+        Scanner scanner = null;
         try {
```
kchilton2: Both of these could be try with resources.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/QueryResultUpdater.java
```diff
@@ -53,32 +54,46 @@
...
-            if(childBindingSet.hasBinding(bindingName)) {
-                final Binding binding = childBindingSet.getBinding(bindingName);
-                queryBindingSet.addBinding(binding);
-            }
+        // Create the Row Key for the result. If the child node groups results, then the key must only contain the Group By variables.
+        final Bytes resultRow;
+
+        final String childNodeId = queryMetadata.getChildNodeId();
+        final boolean isGrouped = childNodeId.startsWith( IncrementalUpdateConstants.AGGREGATION_PREFIX );
+        if(isGrouped) {
```
meiercaleb: Is this necessary?  Can the names of the Projection variables actually differ from the variables that appear in the group by clause?  I thought that the query parser prevented the projection variables from being different than the group by/ aggregation variables.

dao/accumulo.rya/src/main/java/org/apache/rya/accumulo/instance/AccumuloRyaInstanceDetailsRepository.java
kchilton2: Actually, never mind. The Accumulo Scanner interface doesn't implement AutoCloseable.

extras/indexing/src/main/java/org/apache/rya/api/client/accumulo/AccumuloCreatePCJ.java
```diff
@@ -139,16 +138,15 @@ private void updateFluoApp(final String ryaInstance, final String fluoAppName, f
...
                 cd.getInstanceName(),
                 cd.getZookeepers(),
-                fluoAppName);
-
-        // Initialize the PCJ within the Fluo application.
-        final org.apache.rya.indexing.pcj.fluo.api.CreatePcj fluoCreatePcj = new org.apache.rya.indexing.pcj.fluo.api.CreatePcj();
-        fluoCreatePcj.withRyaIntegration(pcjId, pcjStorage, fluoClient, getConnector(), ryaInstance);
+                fluoAppName);) {
+            // Initialize the PCJ within the Fluo application.
+            final org.apache.rya.indexing.pcj.fluo.api.CreatePcj fluoCreatePcj = new org.apache.rya.indexing.pcj.fluo.api.CreatePcj();
```
kchilton2: No. The class name CreatePcj is used twice in the project. This class implements one, which it imports, so this one can't be.

dao/accumulo.rya/src/main/java/org/apache/rya/accumulo/utils/VisibilitySimplifier.java
```diff
@@ -20,26 +20,48 @@
...
+            return vis2;
+        }
+
+        if(vis2.isEmpty()) {
+            return vis1;
+        }
+
+        return simplify("(" + vis1 + ")&(" + vis2 + ")");
+    }
+
```
kchilton2: The whitespace is as I intended, so I'm going to leave it as is.

extras/rya.indexing.pcj/src/main/java/org/apache/rya/indexing/pcj/storage/accumulo/PcjTables.java
```diff
@@ -231,9 +240,10 @@ public PcjMetadata getPcjMetadata(
         checkNotNull(accumuloConn);
         checkNotNull(pcjTableName);
 
+        Scanner scanner = null;
         try {
```
kchilton2: Apparently that interface doesn't extends AutoCloseable.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/VisibilityBindingSetSerde.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Serializes a {@link VisibilityBindingSet} into a {@link Bytes} object.
+     *
+     * @param bindingSet - The binding set that will be serialized. (not null)
+     * @return The serialized object.
+     * @throws Exception A problem was encountered while serializing the object.
+     */
+    public Bytes serialize(final VisibilityBindingSet bindingSet) throws Exception {
+        requireNonNull(bindingSet);
+
```
kchilton2: Yea. There's no reason to write our own serialization for the Value sections, which just hold bytes.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/AggregationResultUpdater.java
```diff
@@ -0,0 +1,583 @@
...
+
+        // Make sure the child binding set has not already been handled.
+        final Bytes childValue = tx.get(childRow, childCol);
+        final VisibilityBindingSet childBindingSet = BS_SERDE.deserialize(childValue);
+
+        log.trace(
+                "Transaction ID: " + tx.getStartTimestamp() + "\n" +
+                "Row Key: " + childRow + "\n" +
+                "Column Key: " + childCol + "\n" +
+                "Binding Set:\n" + childBindingSet + "\n");
```
kchilton2: The column is also being used to fetch the value of the child row.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/AggregationResultUpdater.java
```diff
@@ -0,0 +1,583 @@
...
+/**
+ * Updates the results of an Aggregate node when its child has added a new Binding Set to its results.
+ */
+@DefaultAnnotation(NonNull.class)
+public class AggregationResultUpdater {
+    private static final Logger log = Logger.getLogger(AggregationResultUpdater.class);
+
+    private static final AggregationStateSerDe AGG_STATE_SERDE = new ObjectSerializationAggregationStateSerDe();
+    private static final VisibilityBindingSetSerde BS_SERDE = new VisibilityBindingSetSerde();
+
```
kchilton2: Done.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/AggregationResultUpdater.java
```diff
@@ -0,0 +1,583 @@
...
+     * that is being counted by the {@link AggregationElement}.
+     */
+    public static final class CountFunction implements AggregationFunction {
+        @Override
+        public void update(final AggregationElement aggregation, final AggregationState state, final VisibilityBindingSet childBindingSet) {
+            checkArgument(aggregation.getAggregationType() == AggregationType.COUNT, "The CountFunction only accepts COUNT AggregationElements.");
+
+            // Only add one to the count if the child contains the binding that we are counting.
+            final String aggregatedName = aggregation.getAggregatedBidingName();
+            if(childBindingSet.hasBinding(aggregatedName)) {
```
kchilton2: Done.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/QueryResultUpdater.java
kchilton2: I'm actually not sure. Let me check that.

kchilton2: If I remove that code, all of my aggregation integration tests fail.

kchilton2: Ok, I don't know if this is related, but here was the initial motivation. We're not going to write on top of the existing query result every time if we don't use the group by arguments as the Row Key for the result. So if we're doing the Max function and the first binding set that is written isn't always the max, then you'll have at least two binding sets. We are not guaranteed that the order of the binding set notification will match the order that the binding sets were written, so it's possible for the older result to be exported after the newer result. Which is wrong.

extras/indexing/src/main/java/org/apache/rya/api/client/accumulo/AccumuloDeletePCJ.java
```diff
@@ -118,14 +116,14 @@ private void stopUpdatingPCJ(final String fluoAppName, final String pcjId) {
...
-        final FluoClient fluoClient = new FluoClientFactory().connect(
+        try(final FluoClient fluoClient = new FluoClientFactory().connect(
                 cd.getUsername(),
                 new String(cd.getPassword()),
                 cd.getInstanceName(),
                 cd.getZookeepers(),
-                fluoAppName);
-
-        // Delete the PCJ from the Fluo App.
-        new DeletePcj(1000).deletePcj(fluoClient, pcjId);
```
kchilton2: I'll go do that.

extras/rya.pcj.fluo/pcj.fluo.api/src/main/java/org/apache/rya/indexing/pcj/fluo/api/CreatePcj.java
```diff
@@ -82,6 +82,7 @@
  */
 @DefaultAnnotation(NonNull.class)
 public class CreatePcj {
+    private static final Logger log = Logger.getLogger(CreatePcj.class);
```
kchilton2: Which logging framework are we using? I can update my code at the very least to use that. Is it SLF4J?

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/VisibilityBindingSetSerde.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.rya.indexing.pcj.storage.accumulo.VisibilityBindingSet;
+
+import edu.umd.cs.findbugs.annotations.DefaultAnnotation;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Serializes and deserializes a {@link VisibilityBindingSet} to and from {@link Bytes} objects.
+ */
+@DefaultAnnotation(NonNull.class)
+public class VisibilityBindingSetSerde {
```
kchilton2: I stole the naming convention from Apache Hive.

extras/rya.indexing.pcj/src/main/java/org/apache/rya/indexing/pcj/storage/accumulo/AccumuloPcjSerializer.java
```diff
@@ -112,24 +109,6 @@ public BindingSet convert(byte[] bindingSetBytes, VariableOrder varOrder) throws
...
-     */
-    private static void checkBindingsSubsetOfVarOrder(BindingSet bindingSet, VariableOrder varOrder) throws IllegalArgumentException {
-        checkNotNull(bindingSet);
-        checkNotNull(varOrder);
-
-        Set<String> bindingNames = bindingSet.getBindingNames();
-        List<String> varNames = varOrder.getVariableOrders();
-        checkArgument(varNames.containsAll(bindingNames), "The BindingSet contains a Binding whose name is not part of the VariableOrder.");
-    }
-
```
kchilton2: I don't think it's needed. If a BindingSet has some names that aren't part of the VariableOrder, we can just omit those binidngs from the string that is created by the serializer.

kchilton2: Though I think the logic of this method is actually backwards. It should be checking to ensure the bindingNames contain all of the var names, not the other way around. I could put this method back and fix the logic.

extras/rya.indexing.pcj/src/main/java/org/apache/rya/indexing/pcj/storage/accumulo/BindingSetStringConverter.java
```diff
@@ -58,7 +56,13 @@
...
     @Override
     public String convert(final BindingSet bindingSet, final VariableOrder varOrder) {
-        checkBindingsSubsetOfVarOrder(bindingSet, varOrder);
+        requireNonNull(bindingSet);
+        requireNonNull(varOrder);
+
+        // If the binding set is empty, just return empty string.
+        if(bindingSet.getBindingNames().isEmpty()) {
+            return "";
+        }
```
kchilton2: I don't think I need this change anymore, so I'm removing it.

extras/rya.indexing.pcj/src/main/java/org/apache/rya/indexing/pcj/storage/accumulo/AccumuloPcjSerializer.java
kchilton2: Actually, never mind. It's isn't needed anymore. You're allowed to omit a binding for a variable that is in the variable order because we need that for optional support.

So I'm going to remove this check and just add another unit test.

extras/rya.pcj.fluo/pcj.fluo.api/src/main/java/org/apache/rya/indexing/pcj/fluo/api/CreatePcj.java
isper3at: not sure, but we need to pick one and stick to it.

extras/rya.indexing.pcj/src/main/java/org/apache/rya/indexing/pcj/storage/accumulo/AccumuloPcjSerializer.java
kchilton2: Done.

extras/rya.pcj.fluo/pcj.fluo.api/src/main/java/org/apache/rya/indexing/pcj/fluo/api/CreatePcj.java
kchilton2: Ok, I'm not going to change anything in this review then.

extras/indexing/src/main/java/org/apache/rya/api/client/accumulo/AccumuloDeletePCJ.java
kchilton2: Done.

kchilton2:
I've addressed everything.

extras/rya.pcj.fluo/pcj.fluo.app/src/main/java/org/apache/rya/indexing/pcj/fluo/app/AggregationResultUpdater.java
meiercaleb: Given that the VisibilityBindingSet is available when you call this method, it seems like you are just passing in the Column for logging purposes.

kchilton2: Done. You're still going to need to do a bit of surgery in the future if you want to generalize this problem. I think the Row and Column being passed in make more sense than the observation. The nodes will also have to fetch their own metadata.

PR #159, opened 2017-05-19
F/rya 67 faster integration tests

jwonders:
## Description
Adds a singleton container for the Accumulo MiniCluster to share the MiniCluster across tests (within a project).  Also adds a test rule that can be used to get unique rya instance names and user names within an execution of the tests.

A new fluo instance is still created per test and this is now the dominating contributor to test runtime.  There does not appear to be an obvious way to share the fluo instance since each test will have a separate rya instance and the export observer must be configured up front with the rya instance name.

### Tests
Tests were updated to avoid conflicting rya instance names, user names, etc.  No new tests were added because no functionality was added.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-67)

### Checklist
- [ ] Code Review
- [x] Squash Commits

#### People To Reivew
@meiercaleb @isper3at @pujav65 @amihalik

### DIFF
--- a/dao/accumulo.rya/pom.xml
+++ b/dao/accumulo.rya/pom.xml
@@ -129,6 +129,9 @@ under the License.
                         <artifactId>maven-shade-plugin</artifactId>
                         <executions>
                             <execution>
+                                <goals>
+                                    <goal>shade</goal>
+                                </goals>
                                 <configuration>
                                     <transformers>
                                         <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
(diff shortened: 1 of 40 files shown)


meiercaleb:
Is there any particular reason that you opted to roll your own singleton as opposed to using the Accumulo plugin?  It was our intention to use the plugin to resolve this issue.  However, I noticed in the documentation that the plugin only provides limited configuration capabilities, which could lead to a fairly extensive refactoring of the integration tests.  Was this your reason for not using it?

jwonders:
I described some options in the Jira issue, and the maven plugin was one.  Configurability was not a big driver of the decision.  The downside of the plugin approach is that IDEs do not by default know to use the plugin to start and stop the minicluster.  I think it is possible to configure the major ones to run maven goals before and after tests, but it seems like more work for the developer.  I usually do not like to use singletons, but I think this is a rare situation where a singleton is appropriate.  Note, that it would be possible to introduce a service loading layer to make a choice at runtime between the maven plugin, a lazy-singleton, and even a real accumulo cluster.  This introduces more complexity though so I went with what I deemed to be the simplest approach for both implementation and developer.

PR #162, opened 2017-05-31
RYA-119 Added MongoDB Column Visibility (called Document Visibility).

ejwhite922:
## Description
Re-creating PR that was closed before.

This adds a new field to each document called documentVisibility which
uses a boolean expression to determine if the user can access the document.
The boolean expression is in Disjunctive Normal Formal so that the expression's grouping is simplified enough that MongoDB can run Set operations on it to determine if
the document is viewable.  The expression is stored as an array in MongoDB.

The classes in the package "org.apache.rya.mongodb.document.visibility" are pretty much a copy of the classes that are used by "org.apache.accumulo.core.security.ColumnVisibility" so we don't need to have an Accumulo dependency in a MongoDB component.  If anyone is against that then let me know.

### Tests
Unit Tests

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-119)

### Checklist
- [x] Code Review
- [x] Squash Commits

#### People To Review
@pujav65 
@amihalik 
@isper3at 
@DLotts

### DIFF
--- a/common/rya.api/src/main/java/org/apache/rya/api/persist/RyaDAO.java
+++ b/common/rya.api/src/main/java/org/apache/rya/api/persist/RyaDAO.java
@@ -8,9 +8,9 @@ package org.apache.rya.api.persist;
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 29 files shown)


meiercaleb:
asfbot build

PR #173, opened 2017-06-22
RYA-304 Fixed rya.merger integration tests

ejwhite922:
## Description
rya.merger integration tests were failing and getting incorrect values for the expected query result counts.  Making sure inferencing was passed on to the sail config fixed this.

### Tests
Unit Tests

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-304)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Review
@isper3at 
@DLotts
@meiercaleb

### DIFF
--- a/extras/rya.merger/src/test/java/org/apache/rya/accumulo/mr/merge/RulesetCopyIT.java
+++ b/extras/rya.merger/src/test/java/org/apache/rya/accumulo/mr/merge/RulesetCopyIT.java
@@ -31,6 +31,23 @@ import java.util.Set;
 import org.apache.accumulo.core.client.Connector;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
+import org.apache.rya.accumulo.AccumuloRdfConfiguration;
+import org.apache.rya.accumulo.AccumuloRyaDAO;
+import org.apache.rya.accumulo.mr.MRUtils;
+import org.apache.rya.accumulo.mr.merge.common.InstanceType;
+import org.apache.rya.accumulo.mr.merge.demo.util.DemoUtilities;
+import org.apache.rya.accumulo.mr.merge.demo.util.DemoUtilities.LoggingDetail;
+import org.apache.rya.accumulo.mr.merge.driver.AccumuloDualInstanceDriver;
+import org.apache.rya.accumulo.mr.merge.util.AccumuloRyaUtils;
+import org.apache.rya.accumulo.mr.merge.util.TestUtils;
+import org.apache.rya.api.RdfCloudTripleStoreConfiguration;
+import org.apache.rya.api.domain.RyaStatement;
+import org.apache.rya.api.domain.RyaType;
+import org.apache.rya.api.domain.RyaURI;
+import org.apache.rya.api.persist.RyaDAOException;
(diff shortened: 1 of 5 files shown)


extras/rya.merger/src/test/java/org/apache/rya/accumulo/mr/merge/RulesetCopyIT.java
```diff
@@ -144,32 +146,33 @@ private static RyaType literal(final String lit, final URI type) {
...
+            rulesetTool.shutdown();
+        }
     }
 
     @AfterClass
     public static void tearDownPerClass() throws Exception {
-        if (rulesetTool != null) {
-            rulesetTool.shutdown();
-        }
         accumuloDualInstanceDriver.tearDown();
```
isper3at: so I moved all the tearDowns() into the '@'after  because miniAccumulos were being left open between tests.  Can you verify that it isn't happening?  Otherwise, doing a full teardown between tests did not add a lot of time to the tests.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/244/

extras/rya.merger/src/test/java/org/apache/rya/accumulo/mr/merge/RulesetCopyIT.java
jessehatfield: I can verify that it this helps. Ran tests from the command line ("mvn verify"), monitored the number of org.apache.accumulo.start.Main processes running: seems to vary between 11 (during tests) down to 5 (between tests), where previously it continued to climb up to a max of 31 for me. And the change makes sense: the tool gets initialized in each test (initialization happens in runRulesetCopyTest, which each test calls), so it should be torn down after each test.

ejwhite922: This way caused the tests to run faster and were a bit more reliable. Trying to spin up a new MAC for each test and tear it down nearly doubled the time for running the tests.  It looked like more java processes were opened by spinning up MACs for each test.

ejwhite922: After checking with everyone, I think we're now good with my changes since they teardown the tool's newly created child instance after each test like Andrew's changes.  My PR just keeps the same parent MAC instance up for the class and tears it down at the end of the class.

jessehatfield: Oh, I suppose there are two parts: The AccumuloDualInstanceDriver is initialized once per class, and in turn initializes the parent MAC. But the child MAC is instead initialized when the tool is run (via CopyTool.createChildInstance), which is once per test. Calling rulesetTool.shutdown() will close the child instance (but not the parent). Previously, this shutdown call was only being made once per class, so the extra child MACs weren't being shut down between tests. This PR fixes that by moving that call into the per-test shutdown. This leaves the parent MAC open between tests (hence the minimum 5 processes), but there's only ever one parent MAC (the one that was initialized by the AccumuloDualInstanceDriver) and it does get shut down at the end, so we don't have these extraneous processes. So I believe this PR solves the problem.

PR #174, opened 2017-06-26
RYA-291 Added owl:hasValue inference

jessehatfield:
Given a type associated with a hasValue property restriction: 1) expand queries
for members of the type to also check for anything with the value; and 2) expand
queries for values of that property to check for instances of the type.

## Description
1. Inference engine now stores information about owl:hasValue property restrictions, which relate a property, a value, and a type, such that members of the type are defined to have that value for that property. (https://www.w3.org/TR/2012/REC-owl2-syntax-20121211/#Individual_Value_Restriction) These type/property/value associations can be retrieved by type or by property.
2. HasValueVisitor rewrites queries according to the hasValue restrictions, expanding statement patterns referencing either the type or the property. Class hierarchy is considered, so if something is declared to belong to a subclass of such a type, then its value can be inferred; and if something has the property/value associated with such a type, then its membership in that type and any superclasses can be inferred. The resulting query tree is a union of the hasValue logic and the original statement pattern. Other inference visitors may still transform the original statement pattern.

### Tests
InferenceEngineTest verifies that the correct schema is extracted from the ontology triples; HasValueVisitorTest verifies that query trees are expanded as expected; and InferenceIT verifies expected query results given ontology and instance triples.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-291)

### Checklist
- [x] Code Review
- [x] Squash Commits

#### People To Reivew
@meiercaleb 
@ejwhite922

### DIFF
--- a/sail/src/main/java/org/apache/rya/rdftriplestore/RdfCloudTripleStoreConnection.java
+++ b/sail/src/main/java/org/apache/rya/rdftriplestore/RdfCloudTripleStoreConnection.java
@@ -49,6 +49,7 @@ import org.apache.rya.rdftriplestore.evaluation.QueryJoinSelectOptimizer;
 import org.apache.rya.rdftriplestore.evaluation.RdfCloudTripleStoreEvaluationStatistics;
 import org.apache.rya.rdftriplestore.evaluation.RdfCloudTripleStoreSelectivityEvaluationStatistics;
 import org.apache.rya.rdftriplestore.evaluation.SeparateFilterJoinsVisitor;
+import org.apache.rya.rdftriplestore.inference.HasValueVisitor;
 import org.apache.rya.rdftriplestore.inference.InferenceEngine;
 import org.apache.rya.rdftriplestore.inference.InverseOfVisitor;
 import org.apache.rya.rdftriplestore.inference.PropertyChainVisitor;
@@ -345,6 +346,7 @@ public class RdfCloudTripleStoreConnection extends SailConnectionBase {
                     && this.inferenceEngine != null
                     ) {
                 try {
+                    tupleExpr.visit(new HasValueVisitor(queryConf, inferenceEngine));
                     tupleExpr.visit(new PropertyChainVisitor(queryConf, inferenceEngine));
                     tupleExpr.visit(new TransitivePropertyVisitor(queryConf, inferenceEngine));
                     tupleExpr.visit(new SymmetricPropertyVisitor(queryConf, inferenceEngine));
--- a/sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
+++ b/sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
@@ -1,0 +1,141 @@
+package org.apache.rya.rdftriplestore.inference;
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+
+import org.apache.rya.api.RdfCloudTripleStoreConfiguration;
+import org.apache.rya.api.utils.NullableStatementImpl;
+import org.apache.rya.rdftriplestore.utils.FixedStatementPattern;
+import org.openrdf.model.Resource;
+import org.openrdf.model.URI;
+import org.openrdf.model.Value;
+import org.openrdf.model.vocabulary.OWL;
+import org.openrdf.model.vocabulary.RDF;
+import org.openrdf.query.algebra.StatementPattern;
+import org.openrdf.query.algebra.TupleExpr;
+import org.openrdf.query.algebra.Var;
+
+/**
+ * Expands the query tree to account for any relevant has-value class
+ * expressions in the ontology known to the {@link InferenceEngine}.
+ *
+ * Only operates on {@link StatementPattern} nodes, and only those including a
+ * defined type or defined predicate which is relevant to a has-value
+ * expression in the ontology. When applicable, replaces the node with one or
+ * more nested {@link InferUnion}s, one of whose leaves is the original
+ * StatementPattern.
+ *
+ * A has-value class expression references a specific predicate and a specific
+ * value (object or literal), and represents the set of all individuals having
+ * the specified value for the specified predicate. This has two implications
+ * for inference: 1) If an individual has the specified value for the specified
+ * predicate, then it implicitly belongs to the has-value class expression; and
+ * 2) If an individual belongs to the has-value class expression, then it
+ * implicitly has the specified value for the specified predicate.
+ *
+ * To handle the first case, the visitor expands statement patterns of the form
+ * "?individual rdf:type :class" if the class or any of its subclasses are known
+ * to be has-value class expressions. (Does not apply if the class term
+ * is a variable.) The resulting query tree will match individuals who
+ * implicitly belong to the class expression because they have the specified
+ * value for the specified property, as well as any individuals explicitly
+ * stated to belong to the class.
+ *
+ * To handle the second case, the visitor expands statement patterns of the form
+ * "?individual :predicate ?value" if the predicate is known to be referenced by
+ * any has-value expression. (Does not apply if the predicate term is a
+ * variable.) The resulting query tree will match individuals and values that
+ * can be derived from the individual's membership in a has-value class
+ * expression (which itself may be explicit or derived from membership in a
+ * subclass of the has-value class expression), as well as any individuals and
+ * values explicitly stated.
+ */
+public class HasValueVisitor extends AbstractInferVisitor {
+    /**
+     * Creates a new {@link HasValueVisitor}, which is enabled by default.
+     * @param conf The {@link RdfCloudTripleStoreConfiguration}.
+     * @param inferenceEngine The InferenceEngine containing the relevant ontology.
+     */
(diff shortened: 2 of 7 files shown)


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/246/

sail/src/main/java/org/apache/rya/rdftriplestore/RdfCloudTripleStoreConnection.java
```diff
@@ -345,6 +346,7 @@ protected void commitInternal() throws SailException {
                     && this.inferenceEngine != null
                     ) {
                 try {
```
jessehatfield: Yeah, this is a general issue with Rya's approach to inference (for example, TransitivePropertyVisitor is applied before SubPropertyOfVisitor and SameAsVisitor, and any expansion produced by any of the three will be passed over by the others). In this case, since the original statement is preserved as one branch of the union, there's no conflict as long as the HasValueVisitor is applied first. In cases where there is a conflict, keep in mind that it only manifests when both inference rules in question actually would apply to the same query (e.g. if hasAncestor is transitive, and is a subproperty of hasRelative, and the data says A hasAncestor B has Ancestor C hasAncestor D, I don't believe a query for "?x hasRelative D" will return A). A related limitation is that each visitor only applies once, even if another visitor later produces an expansion that would be relevant. For example, if p1 and p2 are inverse properties, and we have a hasValue condition for p1, queries for p2 won't trigger the hasValue expansion.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    protected void meetSP(StatementPattern node) throws Exception {
+        final Var subjVar = node.getSubjectVar();
+        final Var predVar = node.getPredicateVar();
+        final Var objVar = node.getObjectVar();
+        // We can reason over two types of statement patterns:
+        // { ?var rdf:type :Restriction } and { ?var :property ?value }
+        // Both require defined predicate
```
jessehatfield: In the first case (when the predicate is rdf:type), objVar is some class that can be inferred by a hasValue expression, which would either be the the property restriction itself or any superclass of it. E.g. if we have ":A owl:onProperty :p1 ; owl:hasValue :v1" and ":A rdfs:subClassOf :B" then an objVar with value :A or :B should trigger rewriting. In the second case, where the predicate is something other than rdf:type, we instead check to see if any hasValue expressions involve the predicate, and objVar corresponds to the value variable.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+                        }
+                        currentNode = new InferUnion(currentNode, new InferJoin(relevantValues, valueSP));
+                    }
+                    node.replaceWith(currentNode);
+                }
+            }
+            else {
+                // If the predicate has some hasValue restriction associated with it, then finding
+                // that the object belongs to the appropriate type implies a value.
+                final Map<Resource, Set<Value>> impliedValues = inferenceEngine.getHasValueByProperty(predURI);
```
jessehatfield: All classes with such a restriction, and the values associated with them, and we're trying to match objVar to the value. At this point, the check RDF.TYPE.equals(predURI) failed, which means there isn't any reason to expect objVar to be a class. Instead, we're looking at things like "?subj :p1 ?obj" and saying: If there is a property restriction on p1, involving class C1 and value v1, then this expresses the fact that all members x of :C1 implicitly have the triple (x, :p1, :v1). Therefore, for any x found to belong to :C1, {?subj=x, ?obj=:v1} is a correct solution to the original query. This map from Resources to Values would include an entry mapping C1 to the set {v1}.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+                    final Var valueVar = new Var("v-" + UUID.randomUUID());
+                    TupleExpr currentNode = node.clone();
+                    for (URI property : sufficientValues.keySet()) {
+                        final Var propVar = new Var(property.toString(), property);
+                        final TupleExpr valueSP = new DoNotExpandSP(subjVar, propVar, valueVar);
+                        final FixedStatementPattern relevantValues = new FixedStatementPattern(objVar, propVar, valueVar);
+                        for (Value value : sufficientValues.get(property)) {
+                            relevantValues.statements.add(new NullableStatementImpl(objType, property, value));
+                        }
+                        currentNode = new InferUnion(currentNode, new InferJoin(relevantValues, valueSP));
```
jessehatfield: So the way I understand it, something like ":C1 owl:hasValue :v1 ; owl:onProperty :p1" means: "C1 is the set of individuals who have value v1 for property p1." Therefore, anything with p1=v1 is by definition a member of C1. If we have n such classes C1...Cn, and they all have "C(i) rdfs:subClassOf :Superclass1", then we can say: "The set of individuals with value v1 for property p1 are a subset of Superclass1; the set of individuals with value v2 for property p2 are a subset of Superclass1; ... ; the set of individuals with value vn for property pn are a subset of Superclass1." Therefore we can say that anything with p1=v1 is by definition a member of Superclass1 (regardless of its other properties), anything with p2=v2 is by definition a member of Superclass1 (regardless of its other properties), etc. Here, sufficientValues is supposed to capture those (property, value) combinations that, individually, are each sufficient to determine membership in the (super)class.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+        // { ?var rdf:type :Restriction } and { ?var :property ?value }
+        // Both require defined predicate
+        if (predVar != null && predVar.getValue() != null) {
+            final URI predURI = (URI) predVar.getValue();
+            if (RDF.TYPE.equals(predURI) && objVar != null && objVar.getValue() != null
+                    && objVar.getValue() instanceof Resource) {
+                // If the predicate is rdf:type and the type is specified, check whether it can be
+                // inferred using any hasValue restriction(s)
+                final Resource objType = (Resource) objVar.getValue();
+                final Map<URI, Set<Value>> sufficientValues = inferenceEngine.getHasValueByType(objType);
```
jessehatfield: Yep. Any predicate/value combination that would imply that the subject belongs to that type.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
```diff
@@ -583,4 +625,69 @@ public boolean isSchedule() {
     public void setSchedule(boolean schedule) {
         this.schedule = schedule;
     }
+
+    /**
+     * For a given type, return any properties and values such that owl:hasValue restrictions on
+     * those properties could imply this type. No matter how many restrictions are returned, each
+     * one is considered individually sufficient: if a resource has the property and the value, then
+     * it belongs to the provided type. Takes type hierarchy into account, so the value may imply a
```
jessehatfield: If A were a superclass of B and C, rather than a subclass, then we would consider that individual to be of type A. This method is going to return all the hasValue restrictions on all of a type's subclasses, and since belonging to one subclass is sufficient to determine membership in the superclass, meeting one of those restrictions is also sufficient to determine membership in the superclass. Also keep in mind that hasValue property restrictions don't *require* any data to be there; they actually imply it: "x is a member of B <-> x has prop1=val1". So in your example, if an individual is known to be of type A, then it must also have types B and C, and it must also have prop1=val1 and prop2=val2. If those triples aren't in the data, that's OK, we can infer them (though that would be the other rewriting case and handled by the other method). The clearest  statement of this in the standards is probably [here](https://www.w3.org/TR/owl2-profiles/#Reasoning_in_OWL_2_RL_and_RDF_Graphs_using_Rules), specifically cls-hv1 and cls-hv2 in table 6. (This list of rules is incomplete since it only covers OWL RL languages, but the terms it does cover have their standard meaning.)

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
```diff
@@ -360,7 +362,47 @@ public void refreshGraph() throws InferenceEngineException {
             	  }
             	}
             }
-            
+
+            // Get a set of all property restrictions of any type
+            iter = RyaDAOHelper.query(ryaDAO, null, OWL.ONPROPERTY, null, conf);
```
jessehatfield: Agreed, I'll try and clean this up.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
meiercaleb: So a property restriction on A imposes the same property restriction of B?  Or does it just trigger the rewriting, where all restrictions on B are retrieved from the inference engine (which could be a subset of the restrictions on A)?

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
meiercaleb: Okay.  I see.  I think my confusion here stems from my limited understanding of restrictions.  I was thinking of them as somehow being bound to a class or being a restriction placed on a class instead of as a class in themselves.  Thanks for clearing this up.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/HasValueVisitor.java
jessehatfield: I wouldn't describe it as an imposition; the semantics of the hasValue expression are: "If you have p1=v1, then you're an A (and vice versa)". So if the query is "Get all the A's," then we want to return anything with p1=v1 (objVar here is A, and we ask the inference engine for any predicate/value pairs that entail membership in A). And the semantics of the subclass relationship are: "If you're an A, then you're also a B," so if the query is instead "Get all the B's," then we again want to return anything with p1=v1, in addition to anything that turns out to be a B for some other reason (objVar here is B, and we ask the inference engine for any predicate/value pairs that entail membership in B, including those that entail membership in A because membership in A entails membership in B).

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
meiercaleb: Okay.  Thanks for clarifying.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/259/

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/262/

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/263/

PR #180, opened 2017-07-20
RYA-293 Added owl:unionOf inference

jessehatfield:
If A is the union of B and C, then B and C are subclasses of A.

A owl:unionOf (B C D) means that A is the exact set of nodes that belong to any of B, C, and/or D. Therefore:
1. Any node belonging to B is implied to belong to A, any node belonging to C is implied to belong to A, and any node belonging to node D is implied to belong to A. This is the same implication as B rdfs:subClassOf A, C rdfs:subClassOf A, rdfs:subClassOf A, so we handle this by adding these edges to the subclass graph. (Equivalent to *scm-uni* from [this list of rules](https://www.w3.org/TR/owl2-profiles/#Reasoning_in_OWL_2_RL_and_RDF_Graphs_using_Rules).)

2. Any node belonging to A is implied to belong to at least one of B, C, or D. However, we currently lack the advanced reasoning that would be needed to meaningfully apply this to a query (e.g. using some additional schema about B and C to prove by contradiction that the node must belong to D), so this implication is ignored for now.

(This limitation is consistent with the three OWL 2 profiles: EL and QL do not support unionOf and RL only supports the first case.)

## Description
After creating the initial subClassOf graph, the inference engine then checks for any owl:unionOf triples. For each one, the object should refer to the head of a list and the subject is a type representing the union of all the list elements. Traverse the list and add an edge to the subClassOf graph for every member: since anything belonging to the list element belongs to the union type, we can also say that the list element is a subClassOf the union type.

### Tests
InferenceEngineTest.testUnionOf inserts schema including unions and verifies that the InferenceEngine builds the correct subClassOf graph.
InferenceIT.testUnionQuery inserts schema and ontology, refreshes the InferenceEngine, executes a query relying on union types, and verifies that the expected results are returned.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-293)

### Checklist
- [x] Code Review
- [x] Squash Commits

#### People To Review
@meiercaleb
@isper3at
@ejwhite922

### DIFF
--- a/sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
+++ b/sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
@@ -125,6 +125,7 @@ public class InferenceEngine {
     }
 
     public void refreshGraph() throws InferenceEngineException {
+        ValueFactory vf = ValueFactoryImpl.getInstance();
         try {
             //get all subclassof
             Graph graph = TinkerGraph.open();
@@ -135,6 +136,53 @@ public class InferenceEngine {
                     String edgeName = RDFS.SUBCLASSOF.stringValue();
                     Statement st = iter.next();
                     addStatementEdge(graph, edgeName, st);
+                }
+            } finally {
+                if (iter != null) {
+                    iter.close();
+                }
+            }
+
+            // Add unions to the subclass graph: if c owl:unionOf LIST(c1, c2, ... cn), then any
+            // instances of c1, c2, ... or cn are also instances of c, meaning c is a superclass
+            // of all the rest.
+            // (In principle, an instance of c is likewise implied to be at least one of the other
+            // types, but this fact is ignored for now to avoid nondeterministic reasoning.)
+            iter = RyaDAOHelper.query(ryaDAO, null, OWL.UNIONOF, null, conf);
+            try {
+                while (iter.hasNext()) {
+                    Statement st = iter.next();
+                    Value unionType = st.getSubject();
+                    // Traverse the list of types constituting the union
+                    Value current = st.getObject();
+                    while (current instanceof Resource && !RDF.NIL.equals(current)) {
+                        Resource listNode = (Resource) current;
+                        CloseableIteration<Statement, QueryEvaluationException> listIter = RyaDAOHelper.query(ryaDAO,
+                                listNode, RDF.FIRST, null, conf);
+                        try {
+                            if (listIter.hasNext()) {
+                                Statement firstStatement = listIter.next();
+                                if (firstStatement.getObject() instanceof Resource) {
+                                    Resource subclass = (Resource) firstStatement.getObject();
+                                    Statement subclassStatement = vf.createStatement(subclass, RDFS.SUBCLASSOF, unionType);
+                                    addStatementEdge(graph, RDFS.SUBCLASSOF.stringValue(), subclassStatement);
+                                }
+                            }
+                        } finally {
+                            listIter.close();
+                        }
+                        listIter = RyaDAOHelper.query(ryaDAO, listNode, RDF.REST, null, conf);
+                        try {
+                            if (listIter.hasNext()) {
+                                current = listIter.next().getObject();
+                            }
+                            else {
+                                current = RDF.NIL;
+                            }
+                        } finally {
+                            listIter.close();
+                        }
+                    }
                 }
             } finally {
                 if (iter != null) {
@@ -221,7 +269,6 @@ public class InferenceEngine {
             }
             inverseOfMap = invProp;
             
-            ValueFactory vf = ValueFactoryImpl.getInstance();
             iter = RyaDAOHelper.query(ryaDAO, null, 
             		vf.createURI("http://www.w3.org/2002/07/owl#propertyChainAxiom"),
             		null, conf);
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/272/<h2>Build result: ABORTED</span></h2>[...truncated 6.52 MB...][INFO]                                                                         [INFO] ------------------------------------------------------------------------[INFO] Building Apache Rya PCJ Fluo Demo 3.2.11-incubating-SNAPSHOT[INFO] ------------------------------------------------------------------------[INFO] [INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ rya.pcj.fluo.demo ---[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: java.util.concurrent.ExecutionException: Invalid object ID 12 iota=55[INFO] [INFO] --- apache-rat-plugin:0.11:check (check-licenses) @ rya.pcj.fluo.demo ---[INFO] 51 implicit excludes (use -debug for more details).[INFO] Exclude: **/resources/META-INF/services/**[INFO] 4 resources included (use -debug for more details)[INFO] Rat check: Summary of files. Unapproved: 0 unknown: 0 generated: 0 approved: 4 licence.[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: java.util.concurrent.ExecutionException: Invalid object ID 12 iota=55[INFO] [INFO] --- maven-enforcer-plugin:1.3.1:enforce (enforce-mvn) @ rya.pcj.fluo.demo ---[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: java.util.concurrent.ExecutionException: Invalid object ID 12 iota=55[INFO] [INFO] --- maven-remote-resources-plugin:1.5:process (default) @ rya.pcj.fluo.demo ---Build was abortedchannel stoppedSetting status of ee538a9b113312d29927b0e61bb7396e35e112c5 to FAILURE with url https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/272/ and message: 'FAILURE 'Using context: Jenkins: clean package -Pgeoindexing

pujav65:
looks good -- kicking of a build to merge

pujav65:
hey @jessehatfield , it looks like there are test failures:
https://builds.apache.org/job/incubator-rya-fork/24/
Can you verify that these are false failures?  They do not look like they are related to your pull request.

jessehatfield:
I'm not sure what caused those errors, but I agree they don't appear related. Two are thrown by [KafkaExportITBase.embeddedKafkaTest](https://github.com/apache/incubator-rya/blob/master/extras/rya.pcj.fluo/pcj.fluo.integration/src/test/java/org/apache/rya/indexing/pcj/fluo/KafkaExportITBase.java#L277), whose documentation asserts that this doesn't indicate a problem with Rya itself. I'm not seeing this exact error in any of the other currently failing builds, though many are failing in the same vicinity. I'm less confident about the MongoTemporalIndexerTest, but I haven't managed to reproduce that failure locally.

ejwhite922:
Looks good.  Rebasing with master and pushing should fix the build issue.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/367/

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
```diff
@@ -142,6 +143,53 @@ public void refreshGraph() throws InferenceEngineException {
...
+                                    addStatementEdge(graph, RDFS.SUBCLASSOF.stringValue(), subclassStatement);
+                                }
+                            }
+                        } finally {
+                            listIter.close();
+                        }
+                        listIter = RyaDAOHelper.query(ryaDAO, listNode, RDF.REST, null, conf);
+                        try {
+                            if (listIter.hasNext()) {
+                                current = listIter.next().getObject();
```
meiercaleb: Trying to follow the general logic here:  Each list has an RDF.FIRST and RDF.RESET property, where FIRST is a resource and REST is of type list.  So if the list has more than one element, current is set to the list obtained by the RDF.REST query and we go through the loop again.  Is this how all of the SUBCLASSOF statements are created for the union?

jessehatfield:
So this is a case where a few different schema terms (rdfs:subClassOf, owl:unionOf, and eventually owl:equivalentClass) end up being represented by just one term internally (subclass/superclass relationships). We could plausibly organize by either; my intuition is to use the internal representation, since the internal graph being complete and accurate can matter to the other rules. That would mean pulling the logic introduced here, plus the logic introduced in [https://github.com/apache/incubator-rya/pull/184](PR 184) , plus the logic for rdfs:subClassOf (which 184 incidentally simplifies to a method call anyway) into some "refreshSubClassGraph" method. Thoughts on that approach? Would probably do the same with subPropertyOfGraph for consistency (though it's simpler because there's no equivalent to union).

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
jessehatfield: Yep, a union is given as a linked list so we just walk down adding subclass statements until we get to a node with no rdf:rest or with rdf:rest equal to rdf:nil. If the list is poorly-formed or someone tries to express the union using the wrong collection type ([describes unionOf expression](https://www.w3.org/TR/owl2-rdf-based-semantics/#Semantic_Conditions_for_Boolean_Connectives) | [gives interpretation of sequence](https://www.w3.org/TR/owl2-rdf-based-semantics/#Semantic_Conditions)) then it won't work. In most cases that just means the intended union isn't fully represented, but I suppose if it were somehow a cyclical list, we'd end up in an infinite loop.

PR #194, opened 2017-08-03
RYA-325 Renamed rya.console to rya.shell.

jdasch:
## Description
Renamed rya.console to rya.shell.

### Tests
N/A

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-325)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew
@isper3at 
@meiercaleb


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/348/

extras/pom.xml
```diff
@@ -34,7 +34,7 @@ under the License.
         <module>rya.prospector</module>
         <module>rya.manual</module>
         <module>rya.periodic.service</module>
-        <module>rya.console</module>
+        <module>shell</module>
```
amihalik: agree with Andrew.  Why not put this is rya.shell?

jdasch: It makes filesystem navigation by shell faster and yields shorter, less redundant paths.  If that convention was fully expanded, we would end up having:
`incubator-rya/rya.extras/rya.shell` which is more rya's than I'd like to type.

Does windows still have those long path limitations?

amihalik: okay, I'm fine with this.

PR #197, opened 2017-08-04
RYA-298, RYA-299 Domain/range inference.

jessehatfield:
## Description
InferenceEngine: Added a method to query for and infer domain/range schema, incorporating subclass, subproperty, and inverse property information to compute the closure. Extended graph traversal methods to allow traversal in each direction (findParents has the same behavior as before, but is joined by findChildren, both of which call findConnections with a direction parameter).

DomainRangeVisitor: New class rewrites queries for members of a specific defined type to check for domain and range implications as well. Applies to StatementPatterns of the form <?s a Class>, where Class is not a variable, and produces (if domain and/or range exist for that type) a union which includes the original pattern as well.

RdfCloudTripleStoreConnection: Call the DomainRangeVisitor if inference is enabled. Called at the beginning of the sequence of visitors because the original statement is preserved as one branch of the union, so this won't stop other inference rules from being applied in that branch. However, as with the other visitors, the inferred branches of the unions won't be expanded. This means we won't infer the type by applying multiple rules at once (regardless of order of visitor calls), but we will have a union of alternative ways of inferring the type.

### Tests
Added tests to: Verify that InferenceEngine loads and infers expected schema (taking into account class hierarchy, property graph, and inverse properties); verify that the visitor produces the expected query tree; and verify that a type query returns the expected results given an ontology.

### Links
[RYA-298](https://issues.apache.org/jira/browse/RYA-298)
[RYA-299](https://issues.apache.org/jira/browse/RYA-299)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew
@meiercaleb 
@ejwhite922 
@pujav65
@amihalik

### DIFF
--- a/sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
+++ b/sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
@@ -28,9 +28,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Stack;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.openrdf.model.Resource;
 import org.openrdf.model.Statement;
@@ -73,6 +75,8 @@ public class InferenceEngine {
     private Set<URI> transitivePropertySet;
     private Map<Resource, Map<URI, Value>> hasValueByType;
     private Map<URI, Map<Resource, Value>> hasValueByProperty;
+    private Map<URI, Set<URI>> domainByType;
+    private Map<URI, Set<URI>> rangeByType;
 
     private RyaDAO ryaDAO;
     private RdfCloudTripleStoreConfiguration conf;
@@ -371,6 +375,8 @@ public class InferenceEngine {
             	}
             }
 
+            refreshDomainRange();
+
             refreshPropertyRestrictions();
 
         } catch (QueryEvaluationException e) {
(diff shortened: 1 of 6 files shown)


jessehatfield:
Note that the results of a domain/range query can be huge, and hugely redundant, because: A) there can be many predicates inferring the same type, and we'll query for all of them; and B) even for one predicate, we'll generate a result for each instance of that predicate even if we end up inferring the same resource's type many times. (E.g. in a LUBM dataset, querying for all universities will give you the same solution of :University0 for every triple <?student lubm:degreeFrom :University0>.) Wrapping the query in a Distinct introduces its own computational costs, so the decision to do that is left to the caller.

Some optimizations could be possible if we knew we were starting with a relatively small set of candidates: Asking "does this have the type?" for each candidate could be faster than asking for all the ways of inferring the type. But that optimization is more general than just domain and range queries (and could potentially be generalized even further), so I'd consider something like that outside the scope of this problem.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/355/

sail/src/test/java/org/apache/rya/rdftriplestore/inference/DomainRangeVisitorTest.java
```diff
@@ -0,0 +1,140 @@
...
+        final Set<URI> rangePredicates = new HashSet<>();
+        domainPredicates.add(advisor);
+        domainPredicates.add(takesCourse);
+        rangePredicates.add(advisor);
+        when(inferenceEngine.getPropertiesWithDomain(person)).thenReturn(domainPredicates);
+        when(inferenceEngine.getPropertiesWithRange(person)).thenReturn(rangePredicates);
+        final Var subjVar = new Var("s");
+        final StatementPattern originalSP = new StatementPattern(subjVar, new Var("p", RDF.TYPE), new Var("o", person));
+        final Projection query = new Projection(originalSP, new ProjectionElemList(new ProjectionElem("s", "subject")));
+        new DomainRangeVisitor(conf, inferenceEngine).meet(query);
```
ejwhite922: Should this be query.visit(new DomainRangeVisitor(conf, inferenceEngine))?  Or does it not matter?

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
```diff
@@ -73,6 +74,8 @@
     private Set<URI> transitivePropertySet;
     private Map<Resource, Map<URI, Value>> hasValueByType;
     private Map<URI, Map<Resource, Value>> hasValueByProperty;
+    private Map<URI, Set<URI>> domainByType;
```
ejwhite922: We should probably start using ConcurrentHashMaps for all these maps.  The InferenceEngine is not very thread safe right now.

sail/src/test/java/org/apache/rya/rdftriplestore/inference/DomainRangeVisitorTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.openrdf.query.algebra.ProjectionElemList;
+import org.openrdf.query.algebra.StatementPattern;
+import org.openrdf.query.algebra.TupleExpr;
+import org.openrdf.query.algebra.Union;
+import org.openrdf.query.algebra.Var;
+
+public class DomainRangeVisitorTest {
+    private final AccumuloRdfConfiguration conf = new AccumuloRdfConfiguration();
+    private final ValueFactory vf = new ValueFactoryImpl();
+    private final URI person = vf.createURI("lubm:Person");
```
ejwhite922: These URI's can be made static final

jessehatfield:
Updated. At some point it would be nice to make a pass over the inference engine for thread-safety across both the old and new logic, maybe once the currently-pending extensions are added.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/389/

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
meiercaleb: Can one of you create a ticket for this?  The solution of this is probably not within the scope of this task, but this should be resolved.

sail/src/main/java/org/apache/rya/rdftriplestore/inference/InferenceEngine.java
```diff
@@ -363,13 +367,173 @@ public void refreshGraph() throws InferenceEngineException {
...
+     *  - If p has range c1, and c1 is a subclass of c2, then p also has range c2.
+     * @throws QueryEvaluationException
+     */
+    private void refreshDomainRange() throws QueryEvaluationException {
+        Map<URI, Set<URI>> domainByTypePartial = new ConcurrentHashMap<>();
+        Map<URI, Set<URI>> rangeByTypePartial = new ConcurrentHashMap<>();
+        // First, populate domain and range based on direct domain/range triples.
+        CloseableIteration<Statement, QueryEvaluationException> iter = RyaDAOHelper.query(ryaDAO, null, RDFS.DOMAIN, null, conf);
+        try {
+            while (iter.hasNext()) {
```
meiercaleb: It seems like this map building loop is extremely prevalent in the refresh methods for all of the property restrictions.  Is there any way to pull this logic out into a buildRestrictionMap(...) method that takes in the map that is being updated/created/refreshed, and/or the iterator/property URI used to create the iterator?

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/408/

PR #205, opened 2017-08-14
RYA-343 Fixed rya.api connection issue for PCJ-enabled rya table.

jdasch:
## Description
Added a couple configs that weren't getting set and caused a regression.

Purpose of this PR is to get a fix in place now.  We should probably make a new JIRA to ensure this regression is covered by an IT and to expand PCJDetails to store/retrieve this information.

### Tests
None.

### Links
[Jira: RYA-343](https://issues.apache.org/jira/browse/RYA-343)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew
@amihalik 
@meiercaleb

### DIFF
--- a/common/rya.api/src/main/java/org/apache/rya/api/instance/RyaDetailsToConfiguration.java
+++ b/common/rya.api/src/main/java/org/apache/rya/api/instance/RyaDetailsToConfiguration.java
@@ -20,14 +20,14 @@ package org.apache.rya.api.instance;
 
 import static java.util.Objects.requireNonNull;
 
-import edu.umd.cs.findbugs.annotations.DefaultAnnotation;
-import edu.umd.cs.findbugs.annotations.NonNull;
-
 import org.apache.hadoop.conf.Configuration;
 import org.apache.log4j.Logger;
 import org.apache.rya.api.instance.RyaDetails.PCJIndexDetails;
 
 import com.google.common.base.Optional;
+
+import edu.umd.cs.findbugs.annotations.DefaultAnnotation;
+import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
  * Used to fetch {@link RyaDetails} from a {@link RyaDetailsRepository} and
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/388/

common/rya.api/src/main/java/org/apache/rya/api/instance/RyaDetailsToConfiguration.java
```diff
@@ -53,14 +53,16 @@ public static void addRyaDetailsToConfiguration(final RyaDetails details, final
...
-			conf.set(ConfigurationFields.PCJ_UPDATER_TYPE, "FLUO");
-		} else {
-			checkAndSet(conf, ConfigurationFields.USE_PCJ_UPDATER, false);
-		}
+        final PCJIndexDetails pcjDetails = details.getPCJIndexDetails();
+        if (pcjDetails.isEnabled() && pcjDetails.getFluoDetails().isPresent()) {
+            checkAndSet(conf, ConfigurationFields.USE_PCJ_UPDATER, true);
+            conf.set(ConfigurationFields.FLUO_APP_NAME, pcjDetails.getFluoDetails().get().getUpdateAppName());
+            conf.set(ConfigurationFields.PCJ_UPDATER_TYPE, "FLUO");
+            conf.set(ConfigurationFields.PCJ_STORAGE_TYPE, "ACCUMULO");
```
meiercaleb: So this is probably okay for now.  But this method could take in either an instance of AccumuloRyaInstanceDetails or an instance of MongoRyaInstanceDetails.  So the storage type could be Accumulo or Mongo (note that the Mongo storage type doesn't currently exist in master, but will be merged in as part of a pull request).  Given that there is currently no updater for mongo, there will be no fluo details so the the Accumulo storage type will never be set for a Mongo backed instance.  But this could change once an updater is implemented for Mongo.  The best thing to do is to add a getStorageType() method to the RyaInstanceDetails, and then just pass this into the config so that the storage type aligns with the RyaInstanceDetails type.

common/rya.api/src/main/java/org/apache/rya/api/instance/RyaDetailsToConfiguration.java
```diff
@@ -53,14 +53,16 @@ public static void addRyaDetailsToConfiguration(final RyaDetails details, final
...
+        } else {
+            checkAndSet(conf, ConfigurationFields.USE_PCJ_UPDATER, false);
+            conf.set(ConfigurationFields.PCJ_UPDATER_TYPE, "NO_UPDATE");
```
ejwhite922: I realize it was already like this but make use of the PrecomputedJoinUpdaterType and PrecomputedJoinStorageType enums in this method instead of hard-coded strings.
Replace:
"NO_UPDATE" with PrecomputedJoinUpdaterType.NO_UPDATE.toString(),
"FLUO" with PrecomputedJoinUpdaterType.FLUO.toString(), and
"ACCUMULO" with PrecomputedJoinStorageType.ACCUMULO.toString()

jdasch: I completely agree with you on not using the String literals.  Unfortunately `PrecomputedJoinUpdaterType` is in rya.indexing and `RyaDetailsToConfiguration` is in rya.api, so I can't access that type.  I believe (could be wrong) that Caleb had the action item to create (or reference) a JIRA that would enhance the `PCJIndexDetails` class to have a `PCJIndexDetails.getUpdaterType()` and `PCJIndexDetails.getStorageType()` methods, or perhaps we might go in some other direction with the implementation.

meiercaleb: This issue has been created https://issues.apache.org/jira/browse/RYA-344.

PR #254, opened 2017-12-21
RYA-416 Optionally invoke aggregation pipeline to execute MongoDB queries

jessehatfield:
Provides tools for converting some or all of a SPARQL query to an aggregation pipeline.

## Description
Created AggregationPipelineQueryNode to represent a portion of a query that has been transformed into an aggregation pipeline. Evaluating this node submits the pipeline to the Mongo server and returns a PipelineResultIteration, which translates individual results from Mongo documents into binding sets for use higher in the query or by the caller.

Created a visitor and an optimizer to traverse a query tree and determine whether part or all of the query can be converted into an aggregation pipeline and replaced by the equivalent AggregationPipelineQueryNode.

Added configuration field and methods for enabling/disabling this optimization. Defaults to disabled.

### Tests
AggregationPipelineQueryNodeTest verifies that SPARQL expressions are translated into the correct aggregation operators. SparqlToTransformVisitorTest verifies that SPARQL queries are rewritten by substituting subtrees with equivalent pipelines. PipelineQueryIT verifies that the transformed queries return the expected results.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-416)

### Checklist
- [ ] Code Review
- [x] Squash Commits

#### People To Reivew
@pujav65 
@meiercaleb 
@isper3at 
@kchilton2 
@ejwhite922 
@DLotts

### DIFF
--- a/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
+++ b/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
@@ -1,0 +1,856 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.rya.mongodb.aggregation;
+
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.CONTEXT;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.DOCUMENT_VISIBILITY;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.OBJECT;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.OBJECT_HASH;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.OBJECT_TYPE;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.PREDICATE;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.PREDICATE_HASH;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.STATEMENT_METADATA;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.SUBJECT;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.SUBJECT_HASH;
+import static org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy.TIMESTAMP;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.NavigableSet;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.function.Function;
+
+import org.apache.rya.api.domain.RyaStatement;
+import org.apache.rya.api.domain.RyaType;
+import org.apache.rya.api.domain.RyaURI;
+import org.apache.rya.api.domain.StatementMetadata;
+import org.apache.rya.api.resolver.RdfToRyaConversions;
+import org.apache.rya.mongodb.MongoDbRdfConstants;
+import org.apache.rya.mongodb.dao.MongoDBStorageStrategy;
+import org.apache.rya.mongodb.dao.SimpleMongoDBStorageStrategy;
+import org.apache.rya.mongodb.document.operators.query.ConditionalOperators;
+import org.apache.rya.mongodb.document.visibility.DocumentVisibilityAdapter;
+import org.bson.Document;
+import org.bson.conversions.Bson;
+import org.openrdf.model.Literal;
+import org.openrdf.model.Resource;
+import org.openrdf.model.URI;
+import org.openrdf.model.Value;
+import org.openrdf.model.vocabulary.XMLSchema;
+import org.openrdf.query.BindingSet;
+import org.openrdf.query.QueryEvaluationException;
+import org.openrdf.query.algebra.Compare;
+import org.openrdf.query.algebra.ExtensionElem;
+import org.openrdf.query.algebra.ProjectionElem;
+import org.openrdf.query.algebra.ProjectionElemList;
+import org.openrdf.query.algebra.StatementPattern;
+import org.openrdf.query.algebra.ValueConstant;
+import org.openrdf.query.algebra.ValueExpr;
+import org.openrdf.query.algebra.Var;
+import org.openrdf.query.algebra.evaluation.impl.ExternalSet;
+
+import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import com.mongodb.BasicDBObject;
+import com.mongodb.DBObject;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.model.Aggregates;
+import com.mongodb.client.model.BsonField;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Projections;
+
+import info.aduna.iteration.CloseableIteration;
+
+/**
+ * Represents a portion of a query tree as MongoDB aggregation pipeline. Should
+ * be built bottom-up: start with a statement pattern implemented as a $match
+ * step, then add steps to the pipeline to handle higher levels of the query
+ * tree. Methods are provided to add certain supported query operations to the
+ * end of the internal pipeline. In some cases, specific arguments may be
+ * unsupported, in which case the pipeline is unchanged and the method returns
+ * false.
+ */
+public class AggregationPipelineQueryNode extends ExternalSet {
+    /**
+     * An aggregation result corresponding to a solution should map this key
+     * to an object which itself maps variable names to variable values.
+     */
+    static final String VALUES = "<VALUES>";
+
+    /**
+     * An aggregation result corresponding to a solution should map this key
+     * to an object which itself maps variable names to the corresponding hashes
+     * of their values.
+     */
+    static final String HASHES = "<HASHES>";
+
+    /**
+     * An aggregation result corresponding to a solution should map this key
+     * to an object which itself maps variable names to their datatypes, if any.
+     */
+    static final String TYPES = "<TYPES>";
+
+    private static final String LEVEL = "derivation_level";
+    private static final String[] FIELDS = { VALUES, HASHES, TYPES, LEVEL, TIMESTAMP };
+
+    private static final String JOINED_TRIPLE = "<JOINED_TRIPLE>";
+    private static final String FIELDS_MATCH = "<JOIN_FIELDS_MATCH>";
+
+    private static final MongoDBStorageStrategy<RyaStatement> strategy = new SimpleMongoDBStorageStrategy();
+
+    private static final Bson DEFAULT_TYPE = new Document("$literal", XMLSchema.ANYURI.stringValue());
+    private static final Bson DEFAULT_CONTEXT = new Document("$literal", "");
+    private static final Bson DEFAULT_DV = DocumentVisibilityAdapter.toDBObject(MongoDbRdfConstants.EMPTY_DV);
+    private static final Bson DEFAULT_METADATA = new Document("$literal",
+            StatementMetadata.EMPTY_METADATA.toString());
+
+    private static boolean isValidFieldName(String name) {
+        return !(name == null || name.contains(".") || name.contains("$")
+                || name.equals("_id"));
+    }
+
+    /**
+     * For a given statement pattern, represents a mapping from query variables
+     * to their corresponding parts of matching triples. If necessary, also
+     * substitute variable names including invalid characters with temporary
+     * replacements, while producing a map back to the original names.
+     */
+    private static class StatementVarMapping {
+        private final Map<String, String> varToTripleValue = new HashMap<>();
+        private final Map<String, String> varToTripleHash = new HashMap<>();
+        private final Map<String, String> varToTripleType = new HashMap<>();
+        private final BiMap<String, String> varToOriginalName;
+
+        String valueField(String varName) {
+            return varToTripleValue.get(varName);
+        }
+        String hashField(String varName) {
+            return varToTripleHash.get(varName);
+        }
+        String typeField(String varName) {
+            return varToTripleType.get(varName);
+        }
+
+        Set<String> varNames() {
+            return varToTripleValue.keySet();
+        }
+
+        private String replace(String original) {
+            if (varToOriginalName.containsValue(original)) {
+                return varToOriginalName.inverse().get(original);
+            }
+            else {
+                String replacement = "field-" + UUID.randomUUID();
+                varToOriginalName.put(replacement, original);
+                return replacement;
+            }
+        }
+
+        private String sanitize(String name) {
+            if (varToOriginalName.containsValue(name)) {
+                return varToOriginalName.inverse().get(name);
+            }
+            else if (name != null && !isValidFieldName(name)) {
+                return replace(name);
+            }
+            return name;
+        }
+
+        StatementVarMapping(StatementPattern sp, BiMap<String, String> varToOriginalName) {
+            this.varToOriginalName = varToOriginalName;
+            if (sp.getSubjectVar() != null && !sp.getSubjectVar().hasValue()) {
+                String name = sanitize(sp.getSubjectVar().getName());
+                varToTripleValue.put(name, SUBJECT);
+                varToTripleHash.put(name, SUBJECT_HASH);
+            }
+            if (sp.getPredicateVar() != null && !sp.getPredicateVar().hasValue()) {
+                String name = sanitize(sp.getPredicateVar().getName());
+                varToTripleValue.put(name, PREDICATE);
+                varToTripleHash.put(name, PREDICATE_HASH);
+            }
+            if (sp.getObjectVar() != null && !sp.getObjectVar().hasValue()) {
+                String name = sanitize(sp.getObjectVar().getName());
+                varToTripleValue.put(name, OBJECT);
+                varToTripleHash.put(name, OBJECT_HASH);
+                varToTripleType.put(name, OBJECT_TYPE);
+            }
+            if (sp.getContextVar() != null && !sp.getContextVar().hasValue()) {
+                String name = sanitize(sp.getContextVar().getName());
+                varToTripleValue.put(name, CONTEXT);
+            }
+        }
+
+        Bson getProjectExpression() {
+            return getProjectExpression(new LinkedList<>(), str -> "$" + str);
+        }
+
+        Bson getProjectExpression(Iterable<String> alsoInclude,
+                Function<String, String> getFieldExpr) {
+            Document values = new Document();
+            Document hashes = new Document();
+            Document types = new Document();
+            for (String varName : varNames()) {
+                values.append(varName, getFieldExpr.apply(valueField(varName)));
+                if (varToTripleHash.containsKey(varName)) {
+                    hashes.append(varName, getFieldExpr.apply(hashField(varName)));
+                }
+                if (varToTripleType.containsKey(varName)) {
+                    types.append(varName, getFieldExpr.apply(typeField(varName)));
+                }
+            }
+            for (String varName : alsoInclude) {
+                values.append(varName, 1);
+                hashes.append(varName, 1);
+                types.append(varName, 1);
+            }
+            List<Bson> fields = new LinkedList<>();
+            fields.add(Projections.excludeId());
+            fields.add(Projections.computed(VALUES, values));
+            fields.add(Projections.computed(HASHES, hashes));
+            if (!types.isEmpty()) {
+                fields.add(Projections.computed(TYPES, types));
+            }
+            fields.add(Projections.computed(LEVEL, new Document("$max",
+                    Arrays.asList("$" + LEVEL, getFieldExpr.apply(LEVEL), 0))));
+            fields.add(Projections.computed(TIMESTAMP, new Document("$max",
+                    Arrays.asList("$" + TIMESTAMP, getFieldExpr.apply(TIMESTAMP), 0))));
+            return Projections.fields(fields);
+        }
+    }
+
+    /**
+     * Given a StatementPattern, generate an object representing the arguments
+     * to a "$match" command that will find matching triples.
+     * @param sp The StatementPattern to search for
+     * @param path If given, specify the field that should be matched against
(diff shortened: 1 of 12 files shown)


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/571/

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/572/

kchilton2:
Just a heads up, this is going to conflict with RYA-414 because of changes to MongoDBRdfConfiguration.java.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.hadoop.conf.Configurable;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.log4j.Logger;
+import org.apache.rya.mongodb.MongoDBRdfConfiguration;
+import org.openrdf.query.BindingSet;
+import org.openrdf.query.Dataset;
+import org.openrdf.query.algebra.TupleExpr;
+import org.openrdf.query.algebra.evaluation.QueryOptimizer;
+
+public class AggregationPipelineQueryOptimizer implements QueryOptimizer, Configurable {
```
kchilton2: Docs.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.rya.mongodb.aggregation;
+
+import org.apache.hadoop.conf.Configurable;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.log4j.Logger;
```
kchilton2: We're using slf4j instead of log4j.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
```diff
@@ -0,0 +1,56 @@
...
+import org.openrdf.query.algebra.TupleExpr;
+import org.openrdf.query.algebra.evaluation.QueryOptimizer;
+
+public class AggregationPipelineQueryOptimizer implements QueryOptimizer, Configurable {
+    private Configuration conf;
+    private Logger logger = Logger.getLogger(getClass());
+
+    @Override
+    public void optimize(TupleExpr tupleExpr, Dataset dataset, BindingSet bindings) {
+        if (conf instanceof MongoDBRdfConfiguration) {
```
kchilton2: This will likely change into a StatefulMongoDBRdfConfiguration in 414 (which is a MongoDBRdfConfiguration). Is anything about your optimizer stateful? If so, you don't want to make more than one instance of it, so it should be shared within that config.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/document/util/DocumentVisibilityUtil.java
```diff
@@ -328,4 +328,22 @@ public static boolean doesUserHaveDocumentAccess(final Authorizations authorizat
         }
         return list.toArray(new Object[0]);
     }
+
+    /**
+     * Converts a {@link List} into an array of {@link Object}s.
```
kchilton2: It might be worth mentioning that it flattens lists that are found within inputList into the returned array. What does this do to other data structures though? You'll leave a Set or Map as such in the array? Are other collections just allowed to be in the returned array?

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/dao/SimpleMongoDBStorageStrategy.java
```diff
@@ -63,6 +63,10 @@
     public static final String STATEMENT_METADATA = "statementMetadata";
     public static final String DOCUMENT_VISIBILITY = "documentVisibility";
 
+    public static String hash(String value) {
```
kchilton2: Docs.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/SparqlToPipelineTransformVisitor.java
```diff
@@ -0,0 +1,151 @@
...
+import org.openrdf.query.algebra.Projection;
+import org.openrdf.query.algebra.Reduced;
+import org.openrdf.query.algebra.StatementPattern;
+import org.openrdf.query.algebra.helpers.QueryModelVisitorBase;
+
+import com.mongodb.MongoClient;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
+
+public class SparqlToPipelineTransformVisitor extends QueryModelVisitorBase<Exception> {
```
kchilton2: Docs.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/SparqlToPipelineTransformVisitor.java
```diff
@@ -0,0 +1,151 @@
...
+    private MongoCollection<Document> inputCollection;
+
+    public SparqlToPipelineTransformVisitor(MongoCollection<Document> inputCollection) {
+        this.inputCollection = inputCollection;
```
kchilton2: Is this allowed to be null?

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/SparqlToPipelineTransformVisitor.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public SparqlToPipelineTransformVisitor(MongoDBRdfConfiguration conf) {
+        MongoClient mongo = MongoConnectorFactory.getMongoClient(conf);
```
kchilton2: MongoConnectorFactory no longer exists after RYA-414. You will receive a stateful configuration object that contains the MongoClient.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+            varToOriginalName.put(replacement, original);
+            return replacement;
+        }
+    }
+
+    /**
+     * Create a pipeline based on a StatementPattern.
+     * @param baseSP The leaf node in the query tree.
+     */
+    public AggregationPipelineQueryNode(MongoCollection<Document> collection, StatementPattern baseSP) {
```
kchilton2: The documentation of this seems to be out of date since it accepts a collection.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+        Preconditions.checkNotNull(collection);
```
kchilton2: The checks return the checked object, so this could be 
```
import static java.util.Objects.requireNonNull;
...
this.collection = requireNonNull(collection)
```
Comment applies through out this review.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+                    }
+                }
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
```
kchilton2: Objects.hash( /* stuff here */ );

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
```diff
@@ -0,0 +1,132 @@
...
+     *  result documents to equivalent variable names in the original query.
+     *  Where an entry does not exist for a field, the field name and variable
+     *  name are assumed to be the same.
+     * @param bindings A partial solution. May be empty.
+     */
+    public PipelineResultIteration(AggregateIterable<Document> aggIter,
+            Map<String, String> varToOriginalName,
+            BindingSet bindings) {
+        aggIter.batchSize(BATCH_SIZE);
+        this.cursor = aggIter.iterator();
```
kchilton2: All of these need to be null checked.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
```diff
@@ -0,0 +1,132 @@
...
+        this.varToOriginalName = varToOriginalName;
+        this.bindings = bindings;
+        lookahead();
```
kchilton2: This should be removed. All of your methods perform lookahead(). In general, a constructor shouldn't be invoking methods on something that communicates on a network since it may take a long time to complete when things aren't working properly.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
```diff
@@ -0,0 +1,132 @@
...
+    @Override
+    public BindingSet next() throws QueryEvaluationException {
+        lookahead();
+        BindingSet solution = nextSolution;
+        nextSolution = null;
+        return solution;
+    }
+
+    @Override
+    public void remove() throws QueryEvaluationException {
```
kchilton2: This should throw an unsupported exception. Remove means you're removing the previously returned object from the underlying collection. This doesn't do that.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
```diff
@@ -0,0 +1,132 @@
...
+                    varValue = VF.createLiteral(valueString, VF.createURI(typeString));
+                }
+                Binding existingBinding = bindingSet.getBinding(varName);
+                // If this variable is not already bound, add it.
+                if (existingBinding == null) {
+                    bindingSet.addBinding(varName, varValue);
+                }
+                // If it's bound to something else, the solutions are incompatible.
+                else if (!existingBinding.getValue().equals(varValue)) {
+                    return null;
```
kchilton2: Should this throw an exception? Why does it return null?

kchilton2: This would cause your Iteration to periodically return null which will likely result in a NPE by whoever is using it since that would be pretty unexpected.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+    List<Bson> getPipeline() {
+        return pipeline;
+    }
+
+    /**
+     * Add a join with an individual {@link StatementPattern} to the pipeline.
+     * @param sp The statement pattern to join with
+     * @return true if the join was successfully added to the pipeline.
+     */
+    public boolean joinWith(StatementPattern sp) {
```
kchilton2: null check

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+                    Projections.computed(FIELDS_MATCH, Filters.and(eqTests)),
+                    Projections.include(JOINED_TRIPLE, VALUES, HASHES, TYPES, LEVEL, TIMESTAMP));
+            pipeline.add(Aggregates.project(eqProjectOpts));
+            matchOpts.put(FIELDS_MATCH, true);
+        }
+        // 5. Filter for solutions whose triples match the joined statement
+        //  pattern, and, if applicable, whose additional shared variables
+        //  match the current solution.
+        pipeline.add(Aggregates.match(matchOpts));
+        // 5. Project the results to include variables from the new SP (with
```
kchilton2: 6.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+                    maxValueExpr("$" + TIMESTAMP, getFieldExpr.apply(TIMESTAMP), 0)));
+            return Projections.fields(fields);
+        }
+    }
+
+    /**
+     * Generate a projection expression that evaluates to the maximum of two
+     * fields and a default value.
+     */
+    private static Document maxValueExpr(String field1, String field2, Object defaultValue) {
```
isper3at: why not use mongo's $max?
Add $max to Aggregates?

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
```diff
@@ -0,0 +1,882 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Add a $group step to filter out redundant solutions.
+     * @return True if the distinct operation was successfully appended.
+     */
+    public boolean distinct() {
```
isper3at: there is a mongo function for finding distinct values over fields

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/597/

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
jessehatfield: This method is only called internally, by lookahead(). It returns null when the next solution returned by the pipeline is incompatible with any bindings that were passed into the constructor. (The constructor may have received a non-empty BindingSet if the pipeline query is being evaluated in the context of a specific partial solution.) The lookahead method will use this to loop through pipeline results until it finds one that is compatible with the bindings, or until the pipeline yields no more results. The null value should never make its way to the caller. I didn't use an exception here because it isn't really an exceptional case in that context: we should expect some of the solutions to the pipeline subquery to not match the specific candidate solution.

Skipping over non-matching results client-side might be inefficient, and we could implement this instead by adding a "$match" step for the candidate solution onto the pipeline. But this wouldn't solve the main inefficiency which is that either way we're executing the pipeline subquery for each candidate solution. So the approach I've taken here is just to do something that satisfies the API correctly, and hope that we can structure queries to avoid the inherent inefficiency. (For example, if we had `Join(Join(SP1, SP2), Join(SP3, SP4))`, the visitor as currently implemented would turn that into a join of two pipeline nodes and have to pass binding sets from one into the other, which will be slow regardless of how that's done. But if it's instead parsed as `Join(Join(Join(SP1, SP2), SP3), SP4))` then we can translate the whole thing into one pipeline query, which should be faster.)

I added a test for this iterator which includes the case of a provided BindingSet and incompatible partial solutions, verifying that the expected (non-null) results are returned.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
jessehatfield: Changed, though it looks like we're inconsistent on this.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
jessehatfield: Added.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
jessehatfield: Done.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
jessehatfield: Removed.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
jessehatfield: Agreed, done.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/SparqlToPipelineTransformVisitor.java
jessehatfield: Added.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/SparqlToPipelineTransformVisitor.java
jessehatfield: Shouldn't be, check added.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/dao/SimpleMongoDBStorageStrategy.java
jessehatfield: Added.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/document/util/DocumentVisibilityUtil.java
jessehatfield: Removed this method anyway, turned out to be unnecessary. (I had some incorrect way of converting a Document to a DBObject; if done properly, the existing methods are sufficient.)

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
jessehatfield: Changed to "$max", weird oversight on my part. Also added a specific check to the appropriate test to convince myself it works.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
jessehatfield: Updated.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
jessehatfield: Changed.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
jessehatfield: Fixed.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/601/<h2>Failed Tests: <span class='status-failure'>3</span></h2><h3><a name='incubator-rya-master-with-optionals-pull-requests/org.apache.rya:rya.prospector' /><a href='https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/601/org.apache.rya$rya.prospector/testReport'>incubator-rya-master-with-optionals-pull-requests/org.apache.rya:rya.prospector</a>: <span class='status-failure'>3</span></h3><ul><li><a href='https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/601/org.apache.rya$rya.prospector/testReport/org.apache.rya.prospector.mr/ProspectorTest/testCount/'><strong>org.apache.rya.prospector.mr.ProspectorTest.testCount</strong></a></li><li><a href='https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/601/org.apache.rya$rya.prospector/testReport/org.apache.rya.prospector.service/ProspectorServiceEvalStatsDAOTest/testCount/'><strong>org.apache.rya.prospector.service.ProspectorServiceEvalStatsDAOTest.testCount</strong></a></li><li><a href='https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/601/org.apache.rya$rya.prospector/testReport/org.apache.rya.prospector.service/ProspectorServiceEvalStatsDAOTest/testNoAuthsCount/'><strong>org.apache.rya.prospector.service.ProspectorServiceEvalStatsDAOTest.testNoAuthsCount</strong></a></li></ul>

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/603/

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/611/

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/649/

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
kchilton2: I agree that we're inconsistent within our code, but I think our pom files only ever explicitly pull in slf4j. I think log4j is being used accidentally when it is brought in as a transitive dependency.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
```diff
@@ -0,0 +1,67 @@
...
+                tupleExpr.visit(pipelineVisitor);
+            } catch (Exception e) {
+                logger.error("Error attempting to transform query using the aggregation pipeline", e);
+            }
+        }
+    }
+
+    @Override
+    public void setConf(Configuration conf) {
+        this.conf = conf;
```
kchilton2: This should probably throw an IllegalArgumentException if it is provided a conf object that is not of type StatefulMongoDBRdfConfiguration because the code will not work if that is the case. Throwing an exception make it easier to hunt down that bug.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/PipelineResultIteration.java
kchilton2: Oh I see how that's being used now. Thanks for the explanation.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryNode.java
isper3at: yeah I misunderstood, I though SPARQL DISTINCT was similar to mongo's distinct, or vis versa.  I mean you might be able to use distinct, then do another operation to split up the resulting field into separate documents, but if that's the case this is fine.

dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/aggregation/AggregationPipelineQueryOptimizer.java
jessehatfield: Agreed, fixed.

asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/651/

PR #253, opened 2017-12-19
RYA-413 Fixed how MongoDBRyaDAO closed its internal mongoClient.  Updated AbstractMongoDBRdfConfigurationBuilder config tags.

ejwhite922:
## Description
If the MongoDBRyaDAO is destroyed, its internal mongoClient is not disposed correctly.  This leaves the mongoClient closed which causes problems the next time the MongoDBRyaDAO is created.

The mongoClient is managed by MongoConnectorFactory and needs to be null'ed out to be recreated properly (by calling MongoConnectorFactory.closeMongoClient(); instead of mongoClient.close();)

Some new unit tests were added to test reconstructing the DAO.  They failed with the old way of using mongoClient.close() and pass with the new way of using MongoConnectorFactory.closeMongoClient().

Also, all the static final config tags inside AbstractMongoDBRdfConfigurationBuilder were changed from protected to public (like AbstractAccumuloRdfConfigurationBuilder has) so they could be used inside other tests that make use of the tags.

### Tests
Unit tests

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-413)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Review
@meiercaleb 
@kchilton2 
@jessehatfield 
@isper3at 
@DLotts
@pujav65

PR #263, opened 2018-01-16
RYA-413 Fixed how MongoDBRyaDAO closed its internal mongoClient. Updated AbstractMongoDBRdfConfigurationBuilder config tags.

ejwhite922:
## Description
If the MongoDBRyaDAO is destroyed, its internal mongoClient is not disposed correctly. This leaves the mongoClient closed which causes problems the next time the MongoDBRyaDAO is created.

The mongoClient is managed by MongoConnectorFactory and needs to be null'ed out to be recreated properly (by calling MongoConnectorFactory.closeMongoClient(); instead of mongoClient.close();)

Some new unit tests were added to test reconstructing the DAO. They failed with the old way of using mongoClient.close() and pass with the new way of using MongoConnectorFactory.closeMongoClient().

Also, all the static final config tags inside AbstractMongoDBRdfConfigurationBuilder were changed from protected to public (like AbstractAccumuloRdfConfigurationBuilder has) so they could be used inside other tests that make use of the tags.

NOTE: This is the same as [PR 253](https://github.com/apache/incubator-rya/pull/253) just recreated off of master after some rebasing issues with other PR.

### Tests
Unit tests

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-413)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Review
@meiercaleb
@kchilton2
@jessehatfield
@isper3at
@DLotts
@pujav65

### DIFF
--- a/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/AbstractMongoDBRdfConfigurationBuilder.java
+++ b/dao/mongodb.rya/src/main/java/org/apache/rya/mongodb/AbstractMongoDBRdfConfigurationBuilder.java
@@ -40,23 +40,23 @@ public abstract class AbstractMongoDBRdfConfigurationBuilder<B extends AbstractM
     private boolean useMock = false;
     private String host = "localhost";
     private String port = DEFAULT_MONGO_PORT;
-    protected static final String DEFAULT_MONGO_PORT = "27017";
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/incubator-rya-master-with-optionals-pull-requests/663/

PR #311, opened 2019-07-15
Don't switch on null value, instead switch on valid type, use Map.getOrDefault()

rplst8:
Addresses https://issues.apache.org/jira/browse/RYA-518
Added value to FUNCTION_TYPE enum in IndexingFunctionRegistry
Changed SEARCH_FUNCTIONS.get to getOrDefault with default value so that switch statement doesn't switch on null value.

org.apache.rya.indexing.IndexingFunctionRegistryFreeTextTest

Two tests that toggle the org.apache.rya.indexing.accumulo.ConfigUtils.USE_FREETEXT config variable from true to false.  The tests issue a query that use a SPARQL function (specifically STRLEN) that is the XPath based function fn:string-length

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-518)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew

### DIFF
--- a/extras/indexing/src/test/java/org/apache/rya/indexing/IndexingFunctionRegistryFreeTextTest.java
+++ b/extras/indexing/src/test/java/org/apache/rya/indexing/IndexingFunctionRegistryFreeTextTest.java
@@ -1,0 +1,128 @@
+package org.apache.rya.indexing;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.accumulo.core.client.AccumuloException;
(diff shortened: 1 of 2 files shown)


extras/indexing/src/test/java/org/apache/rya/indexing/IndexingFunctionRegistryFreeTextTest.java
```diff
@@ -0,0 +1,128 @@
+package org.apache.rya.indexing;
```
DLotts: Need Apache header here.  Just copy from another file.

rplst8: Done.

rplst8:
Is there something I need to do to get this merged?

DLotts (commented):
asfbot build

PR #316, opened 2020-07-22
RYA-532 Assign specific interface types for Subject, Predicate, Object and Context and just generally clean up the code.

brushworth:
This pull request is the result of a few months worth of work to evaluate Rya for my organization. I have created this pull request to refactor Rya in small ways and to generally tidy up the code in a way that makes no functional changes. I've tried really hard to only include changes that are uncontroversial and generally applicable. Most of these changes are simply code formatting, import statement reordering and comment fixing.

I've migrated Rya data types such as RyaIRI and RyaType classes so they now inherit from RDF4J types IRI and Value respectively. I've created corresponding Rya classes for all RDF4J types, so that all terminology is now aligned between Rya and RDF4J. I've updated the RdfToRyaConversions and RyaToRdfConversions classes and implemented the required RDF4J methods on Rya types. Note that you still cannot mix and match RDF4J and Rya classes, because methods like equals() and hashCode() will not be true, since Rya types encode extra properties like column qualifiers, visibilities, timestamps, etc. That isn't really any different from how it already is, but it might change in a future pull request.

This pull request is foundational for the future pull requests I will be submitting. This pull request aims to introduce no new features. I will submit new features and fixes one at a time... but building from the code baseline in this PR.

Please note this PR builds upon https://github.com/apache/rya/pull/315 so you need to take that into account when reviewing the diff. When 315 is merged in, this diff will be simplified.

All existing tests remain. I added tests in a few places (e.g. RangeBindingSetEntriesTest) where they were missing. For new code written, such as in RyaToRdfConversions, I have written new tests.

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-532)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Review
@adinancr 
@jdasch


brushworth:
You can see the diff without the noise of the other PR here: https://github.com/brushworth/rya/pull/1

PR #321, opened 2020-12-14
graduation clean-up:removed DISCLAIMER file, removed incubating name,…

adinancr:
graduation clean-up:removed DISCLAIMER file, removed incubating name, updated mailing lists and git repo names in NOTICE, README.md, pom.xml

removed DISCLAIMER file (since Rya is no longer incubating), removed incubating name, updated mailing lists and git repo names in NOTICE, README.md, pom.xml. 
pom.xml still has <version>4.0.1-incubating-SNAPSHOT</version> but I think the release plugin needs to run to change that to 4.0.1 (in all pom.xml files in the repo)

N/A

### Links
[Jira](https://issues.apache.org/jira/browse/RYA-526)

### Checklist
- [ ] Code Review
- [ ] Squash Commits

#### People To Reivew
@DLotts

### DIFF
--- a/DISCLAIMER
+++ b/DISCLAIMER
@@ -1,8 +1,0 @@
-Apache Rya is an effort undergoing incubation at The Apache Software
-Foundation (ASF), sponsored by the Apache Incubator PMC. Incubation is
-required of all newly accepted projects until a further review indicates
-that the infrastructure, communications, and decision making process have
-stabilized in a manner consistent with other successful ASF projects. While
-incubation status is not necessarily a reflection of the completeness or
-stability of the code, it does indicate that the project has yet to be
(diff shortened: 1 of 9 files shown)


DLotts (changes requested):
This is great!  I did a grep and found a few more.  These bits may  not be supported, what do you think? 
./extras/rya.manual/src/site/markdown/shell.md
./extras/rya.manual/src/site/markdown/rya-streams.md
./extras/rya.manual/src/site/markdown/rya-streams.md

adinancr:
Made one more commit on the branch to replace the actual version with the string "<version>" in the manual, so it would not have to change for each release

DLotts (approved):
Thank you!

